PR #28, opened 2017-04-07
Austin/network

AAAstorga:
Wanting to merge what I have now for networking so Monica can integrate her views. Look in Network.swift for examples. Since we are using static data, code is simple for now.

### DIFF
--- a/TCAT/Model/Route.swift
+++ b/TCAT/Model/Route.swift
@@ -76,6 +76,7 @@ class Route: NSObject, JSONDecodable {
     
     required init(json: JSON) throws {
         super.init()
+        print(json)
         departureTime = timeToDate(time: json["departureTime"].stringValue)
         arrivalTime = timeToDate(time: json["arrivalTime"].stringValue)
         directions = directionJSON(json:json["directions"].array!)
--- a/TCAT/Model/Route.swift
+++ b/TCAT/Model/Route.swift
@@ -72,17 +72,14 @@ class Route: NSObject, JSONDecodable {
     var travelDistance: Double = 0.0 //of first stop
     var lastStopTime: Date = Date()// the critical last time a bus route runs
     
-   
-    
     required init(json: JSON) throws {
         super.init()
-        print(json)
-        departureTime = timeToDate(time: json["departureTime"].stringValue)
-        arrivalTime = timeToDate(time: json["arrivalTime"].stringValue)
+        departureTime = Time.date(from: json["departureTime"].stringValue)
+        arrivalTime = Time.date(from: json["arrivalTime"].stringValue)
         directions = directionJSON(json:json["directions"].array!)
         mainStops = json["stopNames"].arrayObject as! [String]
         mainStopsNums = json["stopNumbers"].arrayObject as! [Int]
-        travelDistance = (directions[0] as! WalkDirection).travelDistance
+        travelDistance = directions[0] is WalkDirection ? (directions[0] as! WalkDirection).travelDistance : 0.0
         lastStopTime = Date()
     }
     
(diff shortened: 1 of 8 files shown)


TCAT/Model/Route.swift
```diff
@@ -62,12 +65,26 @@ class Route: NSObject {
...
-    var lastStopTime: Date // the critical last time a bus route runs
+    var mainStopsNums: [Int] = [Int]()//-1 for pins
+    var travelDistance: Double = 0.0 //of first stop
+    var lastStopTime: Date = Date()// the critical last time a bus route runs
+    
+   
+    
+    required init(json: JSON) throws {
+        super.init()
+        print(json)
```
mattbarker016: Maybe comment these out when merging?

TCAT/Model/Route.swift
```diff
@@ -62,12 +65,26 @@ class Route: NSObject {
...
+        departureTime = timeToDate(time: json["departureTime"].stringValue)
+        arrivalTime = timeToDate(time: json["arrivalTime"].stringValue)
+        directions = directionJSON(json:json["directions"].array!)
+        mainStops = json["stopNames"].arrayObject as! [String]
+        mainStopsNums = json["stopNumbers"].arrayObject as! [Int]
+        travelDistance = (directions[0] as! WalkDirection).travelDistance
```
mattbarker016: Is this a safe assumption? What if there's other walking directions

TCAT/Model/Route.swift
```diff
@@ -78,4 +95,52 @@ class Route: NSObject {
...
+    func directionJSON(json: [JSON]) -> [Direction] {
+        var directionArray = [Direction]()
+        for direction in json {
+            let directionType = direction["directionType"].stringValue
+            if directionType == "walk" {
+                directionArray.append(walkDirection(json: direction))
+            } else if directionType == "depart" {
+                directionArray.append(departDirection(json: direction))
+            } else {
+                directionArray.append(arriveDirection(json: direction))
```
mattbarker016: I would do 3 else if's and an else to handle errors

TCAT/Model/Route.swift
```diff
@@ -78,4 +95,52 @@ class Route: NSObject {
...
+        let time = timeToDate(time: json["time"].stringValue)
+        let place = json["place"].stringValue
+        let location = CLLocation(latitude: json["location"][0].doubleValue, longitude: json["location"][1].doubleValue)
+        return ArriveDirection(time: time, place: place, location: location)
+        
+    }
+    
+    func timeToDate(time: String) -> Date {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = "hh:mm a"
```
mattbarker016: there is a function in Time.swift that does this string(from time). Switch to it so we can stay consistent.

mattbarker016 (commented):
Few stylistic things, plus travelDistance clarification, but switch to the Time.swift function for sure

TCAT/Model/Route.swift
AAAstorga: Actually no it isn't a safe assumption. What does travelDistance mean? The spec in the Route.swift file was a little confusing

mattbarker016: I think it was supposed to be the "25 minutes" for the entire route in the summary view, but I actually just calculate it from arrival and departure time. We can probably get rid of the variable as a whole

Edit: Oh wait it's travel distance. I think Monica uses it for the 0.1 mi away on the first bus stop. Can you check with her that she uses that variable for that reason?

AAAstorga:
012a36b

PR #60, opened 2017-10-22
Austin/fixes

AAAstorga:
Fixed search bar bugs and took out 'Get There Now'. I also changed the pins to the proper colors


TCAT/Controllers/SearchResultsTableViewController.swift
```diff
@@ -54,10 +54,16 @@ class SearchResultsTableViewController: UITableViewController, UISearchBarDelega
     convenience init() {
         self.init(style: .grouped)
     }
-    
+
+    override func viewWillAppear(_ animated: Bool) {
+        if #available(iOS 11.0, *) {
+            searchBar?.heightAnchor.constraint(equalToConstant: 44).isActive = true
```
jhseung: I don't think setting height anchor is necessary.

PR #63, opened 2017-10-29
Monica/ui fixes

myo3:


TCAT/Model/Route.swift
```diff
@@ -144,14 +147,17 @@ class Route: NSObject, JSONDecodable {
...
-     *      remove the first routeSummaryObject
+     *      remove the first routeSummaryObject and update the time
      *  Else (the first routeSummaryObject is a bus stop), so simply update the name to the name the user searched for
      */
     func updateStartingDestination(_ place: Place) {
         if let firstRouteSummaryObject = routeSummary.first {
             if firstRouteSummaryObject.name == "Start" {
                 if (place is PlaceResult || (place is BusStop && place.name == "Current Location")) && routeSummary.count > 2 {
                     routeSummary.remove(at: 0)
+                    departureTime = (routeSummary.first?.time)!
```
AAAstorga, line 158: Is this force unwrap ok? Will it always always work?

AAAstorga (approved):
I left a comment, but besides that it looks great!

TCAT/Model/Route.swift
myo3: yes b/c i check `routeSummary.count > 2` in if statement so `routeSummary.first?` will never be nil and `time` is not an optional property in `RouteSummaryObject`

myo3: but agree this not good style >.<

PR #67, opened 2017-11-05
Matt/route detail

mattbarker016:
I already merged master with @myo3 work


TCAT/Controllers/RouteDetailViewController.swift
```diff
@@ -479,7 +488,7 @@ class RouteDetailViewController: UIViewController, GMSMapViewDelegate, CLLocatio
         // Place and format bottom summary label
         let summaryBottomLabel = UILabel()
         if let totalTime = Time.dateComponents(from: route.departureTime, to: route.arrivalTime).minute {
-            summaryBottomLabel.text = "Trip Duration: \(abs(totalTime)) minutes"
+            summaryBottomLabel.text = "Trip Duration: \(abs(totalTime)) minute\(totalTime == 1 ? "" : "s")"
```
AAAstorga, line 491: thanks for determining between 1 minute and 2 minutes. I dislike when apps say 1 minutes

TCAT/Model/BusLocation.swift
```diff
@@ -28,6 +28,8 @@ class BusLocation: NSObject, NSCoding {
     var tripID: Int = 0
     var vehicleID: Int = 0
     
+    var _iconView: UIView? = nil
```
AAAstorga: why does this have _?

TCAT/Model/BusPath.swift
```diff
@@ -17,56 +17,51 @@ enum PathType: String {
 
 class Path: GMSPolyline {
     
+    var waypoints: [Waypoint] = []
+    var traveledPolyline: GMSPolyline = GMSPolyline()
+    var color: UIColor = .clear
+    
+    init(waypoints: [Waypoint]) {
```
AAAstorga, line 24: remove empty lines

AAAstorga (approved):
small comments, but the rest looks good!

TCAT/Model/BusLocation.swift
mattbarker016: I don't want it to be accessed. That's the syntax for "hidden" variables. I should just make it private https://forums.developer.apple.com/thread/54256

TCAT/Controllers/RouteDetailViewController.swift
mattbarker016: 💙

PR #77, opened 2017-11-20
matt/bug_fixes

mattbarker016:


TCAT/Controllers/RouteDetailViewController.swift
```diff
@@ -95,7 +95,7 @@ class RouteDetailViewController: UIViewController, GMSMapViewDelegate, CLLocatio
                 
                 var waypoints: [Waypoint] = []
                 for point in path {
-                    var type: WaypointType = .none
+                    var type: WaypointType = .walking
                     if direction == self.directions.first && point == path.first! {
```
AAAstorga, line 99: Don't assume first exists. do an if let please

TCAT/Controllers/RouteDetailViewController.swift
```diff
@@ -413,15 +397,16 @@ class RouteDetailViewController: UIViewController, GMSMapViewDelegate, CLLocatio
 
                 UIView.animate(withDuration: 1.0, delay: 0, options: .curveEaseInOut, animations: {
                     newBus.userData = bus
-                    (newBus.iconView as? BusLocationView)?.setBearing(bus.heading, start: existingBus!.position, end: busCoords)
+                    // (newBus.iconView as? BusLocationView)?.setBearing(bus.heading, start: existingBus!.position, end: busCoords)
```
AAAstorga, line 400: can you delete this?

TCAT/Controllers/RouteDetailViewController.swift
```diff
@@ -458,6 +443,21 @@ class RouteDetailViewController: UIViewController, GMSMapViewDelegate, CLLocatio
...
+    func setIndex(of marker: GMSMarker, with waypoint: Waypoint) {
+        marker.zIndex = {
+            switch waypoint.wpType {
+            case .bus: return 1
+            case .walk: return 1
+            case .origin: return 3
+            case .destination: return 3
+            case .stop: return 1
+            case .walking: return 0
+            case .bussing: return 999
```
AAAstorga: explain why you return 999 please

AAAstorga (approved):
after doing the fixes, merge it on in

TCAT/Controllers/RouteDetailViewController.swift
mattbarker016: I toggle between `bearing` and `betterBearing`, not sure why you don't have` betterBearing`

TCAT/Controllers/RouteDetailViewController.swift
mattbarker016: The random high number worked, it's just a high constant

TCAT/Controllers/RouteDetailViewController.swift
mattbarker016: If the for loop fires, there must be at least one element. That's how you get EXCITED in Swift properly

PR #84, opened 2017-12-21
Disappearing Back Button Fix

mattbarker016:


mattbarker016:
Also added way to remove buses no longer being live tracked. Some codebase quality improvements, etc.

TCAT/Views/BusLocationView.swift
```diff
@@ -98,16 +96,18 @@ class BusLocationView: UIView {
             }
         }
         
+        let newDegrees = Double(degrees) - self.currentBearing
```
myo3: this is kind of nit-picky but do you need the `self.`?

myo3:
looks good to me!

TCAT/Views/BusLocationView.swift
mattbarker016: I look putting a little bit of my`self` into my code

mattbarker016: It was a leftover from being in the `UIView.animate` closure

myo3: lol okay then yeah `self` is necessary

myo3: edit: misread your comment but cool 👍

PR #85, opened 2018-01-06
add register logging

xaanimus:
Add Register logging code. I didn't remove the Crashlytics code because there's currently a bug in the Register server.


TCAT.xcodeproj/project.pbxproj
```diff
@@ -525,7 +530,7 @@
 				TargetAttributes = {
 					449A7C741D80D0E80019300C = {
 						CreatedOnToolsVersion = 7.3.1;
-						DevelopmentTeam = ZGMCXU7X3U;
+						DevelopmentTeam = TD73A242XE;
```
mattbarker016: Keep this to the one we had it. It might be fine, but change it back just in case

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -324,6 +325,10 @@ class RouteOptionsViewController: UIViewController, UITableViewDelegate, UITable
                     self.routeResults.reloadData()
                 })
                 
+                let event = DestinationSearchedEventPayload(destination: (self.searchTo?.name)!,
```
mattbarker016: can we throw a default optional on this in case it isn't there?

mattbarker016 (commented):
good stuff, just fix quick changes. It would be helpful if there were some general comments so we can continue adding more events

TCAT/Controllers/RouteOptionsViewController.swift
xaanimus: 👍

xaanimus:
Cool

PR #86, opened 2018-01-08
Make label multi-line for long stop names

myo3:


TCAT/Views/RouteDiagram.swift
```diff
@@ -109,7 +109,11 @@ class RouteDiagram: UIView {
     // MARK: Get data from route ojbect
     
     private func getStopNameLabel() -> UILabel {
-        let stopNameLabel = UILabel(frame: CGRect(x: 0, y: 0, width: 200, height: 20))
+        let yPos: CGFloat = 101
```
mattbarker016, line 112: ahhhh such a non even number

it's coo tho

mattbarker016 (commented):
neato

PR #87, opened 2018-01-10
fixed constraint issues

AAAstorga:
Also @mattbarker016 I could not get it to build no matter what I did without commenting out the one line. Does anyone else have this problem?


myo3:
I deleted the `config.json`from the pbxproj and re-added it and that made it build

PR #89, opened 2018-02-25
Monica/remove route summary

myo3:


TCAT/Model/Route.swift
```diff
@@ -135,86 +128,11 @@ class Route: NSObject, JSONDecodable {
...
-                    }
-                }
-            }
-        }
-    }
-    
     func isWalkingRoute() -> Bool {
-        for routeSummaryObj in routeSummary {
-            if routeSummaryObj.nextDirection != .walk {
+        for direction in directions {
```
mattbarker016: `let isWalkingRoute = route.directions.reduce(true) { $0 && $1.type == .walk }`
So Swift. So functional.

TCAT/Model/Route.swift
```diff
@@ -224,38 +142,20 @@ class Route: NSObject, JSONDecodable {
...
-                if lastRouteSummaryObject.name == "End" && routeSummary.count > 2 && routeSummary[secondToLastStop].name.lowercased() == place.name.lowercased() {
-                    routeSummary.removeLast()
-                    routeSummary.last?.removeNextDirection() // remove 2nd-to-last stop next direction (so it behaves like ending stop)
-                } else if(lastRouteSummaryObject.name == "End") {
-                    lastRouteSummaryObject.updateName(from: place)
-                }
-            }
-        }
+        let numberOfMetersInMile = 1609.34
+        let distanceInMeters = fromLocation.distance(from: endLocation)
```
mattbarker016, line 151: I'd like to centralize conversions if possible, hmm

mattbarker016 (commented):
Nothing technically *has* to be changed 😛

TCAT/Model/Route.swift
myo3: tru tru

TCAT/Model/Route.swift
myo3: yeah maybe we can think of a class later to put that in or make it an extension of double?

PR #90, opened 2018-02-28
v1 Bug Fixes

mattbarker016:


mattbarker016:
There's a lot, but the biggest changes are with the summaryView and Network. A lot of cleaning everywhere else. I also will need to make one last commit to switch variables / documentation to miles, since we'll now receive that from backend.

TCAT/Controllers/HomeViewController.swift
```diff
@@ -386,7 +386,7 @@ class HomeViewController: UIViewController, UITableViewDelegate, UITableViewData
         navigationItem.rightBarButtonItem = nil
         //Crashlytics Answers
         Answers.searchBarTappedInHome()
-        RegisterSession.shared?.logEvent(event: SearchBarTappedEventPayload(location: .home).toEvent())
+        let _ = RegisterSession.shared?.logEvent(event: SearchBarTappedEventPayload(location: .home).toEvent())
```
myo3: is the `let _` if you don't use the value?

mattbarker016: Yeah, there are warnings if you don't have it. I don't know much about utilizing Register, have been meaning to dig into it

TCAT/Controllers/RouteDetail+DrawerViewController.swift
```diff
@@ -301,65 +275,91 @@ class RouteDetailDrawerViewController: UIViewController, UITableViewDataSource,
...
     }
     
+    /// Toggle the cell expansion at the indexPath
+    func toggleCellExpansion(at indexPath: IndexPath?) {
+        
+        guard
+            let indexPath = indexPath,
+            let cell = tableView.cellForRow(at: indexPath) as? LargeDetailTableViewCell
+        else {
+            return
```
myo3, line 310: maybe print an error msg if this guard let is not supposed to fail

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -42,7 +42,7 @@ class RouteOptionsViewController: UIViewController, UITableViewDelegate, UITable
     // MARK: View vars
 
     var routeSelection: RouteSelectionView!
-    var datePickerView: DatepickerView!
+    var datePickerView: DatePickerView!
```
myo3, line 45: lol fineee it can be `DatePicker` 😛

TCAT/Cells/LargeDetailTableViewCell.swift
```diff
@@ -40,7 +40,7 @@ class LargeDetailTableViewCell: UITableViewCell {
         titleLabel.font = UIFont.systemFont(ofSize: UIFont.systemFontSize)
         titleLabel.lineBreakMode = .byWordWrapping
         titleLabel.textColor = .primaryTextColor
-        titleLabel.text = "Board"
+        titleLabel.text = direction != nil && direction.type == .transfer ? "Bus becomes" : "Board"
```
myo3, line 43: jw when would a direction be nil? i assume every cell would be associated with a direction

TCAT/Cells/RouteTableViewCell.swift
```diff
@@ -109,7 +109,7 @@ class RouteTableViewCell: UITableViewCell {
     
     // MARK: Set Data
         
-    func setRouteData(){
+    func setRouteData() {
```
myo3, line 112: good catch! seems i always forget that space ^^;

TCAT/Controllers/HomeViewController.swift
```diff
@@ -384,13 +387,11 @@ class HomeViewController: UIViewController, UITableViewDelegate, UITableViewData
...
         searchBar.placeholder = nil
         navigationItem.rightBarButtonItem = nil
-        //Crashlytics Answers
-        Answers.searchBarTappedInHome()
         let _ = RegisterSession.shared?.logEvent(event: SearchBarTappedEventPayload(location: .home).toEvent())
     }
     
     func searchBarCancelButtonClicked(_ searchBar: UISearchBar) {
-        searchBar.placeholder = "Search (e.g Balch Hall, 312 College Ave)"
+        searchBar.placeholder = "Search (e.g. Balch Hall, 312 College Ave)"
```
myo3, line 394: punctuation on point 👌

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -227,97 +221,129 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
...
-
+            
+        // Otherwise, add bus to map
+        else {
+            
+            let marker = GMSMarker(position: busCoords)
+            (bus.iconView as? BusLocationView)?.setBearing(bus.heading)
+            marker.iconView = bus.iconView
+            // marker.appearAnimation = .pop
+            // marker.tracksViewChanges = true
```
myo3: do u need the commented out code? ow i would delete it

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -58,13 +58,14 @@ class RouteOptionsViewController: UIViewController, UITableViewDelegate, UITable
 
     // MARK: Reachability vars
 
-    let reachability: Reachability? = Reachability(hostname: Network.source)
+    let reachability: Reachability? = Reachability(hostname: Network.ipAddress)
+    
     var banner: StatusBarNotificationBanner = {
-        let banner = StatusBarNotificationBanner(title: "No Internet Connection", style: .danger)
+        let banner = StatusBarNotificationBanner(title: "No internet connection. Retrying...", style: .danger)
```
myo3, line 64: Does it retry tho? i remember bef austin had a retry button on his view so i assumed reachability doesn't retry the network connection. Maybe test to make sure that if get no internet connection banner but then you connect to internet, the reachability banner disappears

TCAT/Model/BusLocation.swift
```diff
@@ -25,7 +37,7 @@ class BusLocation: NSObject, NSCoding {
     var routeID: String = ""
     var runID: Int = 0
     var speed: Int = 0
-    var tripID: Int = 0
+    var tripID: String = ""
```
myo3: will austin return all these instance vars? if so i'd consider not having a default values for the instance vars and create an initializer that takes in all these class vars so we don't have a case of accidentally using the nonsensical default values. Or make the instance vars optional if we're not guaranteed to get all this data from the backend.

TCAT/Model/Direction.swift
```diff
@@ -83,70 +64,90 @@ class Direction: NSObject, NSCopying {
...
-            switch json["type"].stringValue {
-            case "walk" : return .walk
-            case "depart" : return .depart
-            default : return .other
-            }
-        }()
+        self.init()
         
         name = json["name"].stringValue
+        type = json["type"].stringValue == "depart" ? .depart : .walk
```
myo3: being kind of picky but lowercase the string value just in case

TCAT/Utilities/Network.swift
```diff
@@ -155,49 +159,46 @@ class AllBusStops: JSONDecodable {
...
-
-        }
-
-        return allLocationsArray
+    func parseBusLocation(json: JSON) -> BusLocation {
+        
+        let routeID = json["routeID"].stringValue
+        let busLocation = BusLocation(routeID: routeID)
+        busLocation.dataType = {
+            switch json["case"] {
```
myo3: shouldn't there be a `.stringValue`?

TCAT/Cells/LargeDetailTableViewCell.swift
mattbarker016: Yeah it's a bit weird but this stuff is accessed on init before Direction is called

TCAT/Controllers/HomeViewController.swift
mattbarker016: What a through PR

TCAT/Controllers/RouteOptionsViewController.swift
mattbarker016: Mihir suggested it to show we're always trying. Theoretically it should go away once it changes with Reachability

TCAT/Model/Direction.swift
mattbarker016: It's in the REST doc that it should be that though, but I can

TCAT/Utilities/Network.swift
mattbarker016: What do you mean?

TCAT/Model/BusLocation.swift
mattbarker016: Oh so the way I have this set up is that `init` takes one parameter, and I set the rest my doing `newObject.parameter = value`. I didn't want a crazy long `init(...)` function that I have to keep changing and remembering order and such

TCAT/Controllers/RouteDetail+ContentViewController.swift
mattbarker016: Figured out a solution!

TCAT/Controllers/RouteOptionsViewController.swift
myo3: okay if does go away when we reconnect to internet then it should be okay. don't want to mislead user

TCAT/Model/BusLocation.swift
myo3: i think it's better to have a crazy long init function b/c it indicates a flow. Plus it guarantees that at initialization these variables are going to be filled out and no one's going to accidentally to forgot to fill them out b/c they're in the initializer

TCAT/Model/Direction.swift
myo3: yeah just in case. prob never happen but who knows

TCAT/Utilities/Network.swift
myo3: shouldn't it be `json["case"].stringValue`?

TCAT/Model/BusLocation.swift
mattbarker016: Not a bad point

TCAT/Utilities/Network.swift
mattbarker016: yes!

mattbarker016:
Sound off when good to go! 

Update: Oh right just the BusLocation init

myo3:
what about the BusLocation.swift initializer?

Besides that i think its good

mattbarker016:
Good to go?

TCAT/Model/BusLocation.swift
```diff
@@ -42,15 +42,45 @@ class BusLocation: NSObject, NSCoding {
```
myo3, line 42: Almost. Just n2 remove the default values from the variables that are in the initializer

PR #92, opened 2018-03-07
Monica/ui updats

myo3:
Apologies about the typo. I updated routeOptionsVC to latest UI,<s>`except for transfers and I need to work on distance label running off.`</s>

Implemented transfers and stopped distance label from running off


TCAT/Assets.xcassets/live.imageset/Contents.json
```diff
@@ -0,0 +1,26 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "live.png",
```
mattbarker016, line 5: I added a color parameter in the `LiveIndicator` view. I'll have to add an enum or size parameters as well. We should stick to one use of this throughout so we're consistent with the look and animation. Although if this should be static and different, so be it. I'll ask Mihir

TCAT/Cells/RouteTableViewCell.swift
```diff
@@ -14,221 +14,314 @@ protocol TravelDistanceDelegate: NSObjectProtocol {
...
+                .perform(withSuccess: { (result) in
+                    if let busLocation = result.busLocation {
+                        
+                        switch busLocation.dataType {
+                            
+                        case .validData:
+                            self.setLiveElements(withStartTime: direction.startTime, withDelay: busLocation.delay)
+                            
+                        default:
+                           self.setLiveElements(withStartTime: nil, withDelay: nil)
```
mattbarker016, line 158: a lot of similar code for many cases. Could this be combined, or put in a completion function to not duplicate code?

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -360,6 +358,15 @@ class RouteOptionsViewController: UIViewController, UITableViewDelegate, UITable
 
     }
 
+    // to test walking route
+    private func transformToWalkingRoute(_ route: Route) -> Route {
```
mattbarker016: Why do we need to test this?

TCAT/Model/Route.swift
```diff
@@ -177,12 +177,32 @@ class Route: NSObject, JSONDecodable {
...
     
-    // MARK: Process raw routes
+    // MARK: Process routes
     
     func isWalkingRoute() -> Bool {
         return directions.reduce(true) { $0 && $1.type == .walk }
     }
     
+    func getFirstDepartDirection() -> Direction? {
+        for direction in directions {
```
mattbarker016: small detail, but can use functional programming to return an optional in one line. 

`return directions.first { $0.type == .depart }`

Change condition and ordering as needed.

TCAT/Views/WalkWithDistanceIcon.swift
```diff
@@ -0,0 +1,45 @@
...
+    // MARK: View vars
+    
+    var walkIcon: UIImageView
+    var travelDistanceLabel: UILabel
+    
+    // MARK: Spacing vars
+    
+    let walkIconAndDistanceLabelVerticalSpace: CGFloat = 2.0
+
+    init(walkIcon: UIImageView, travelDistanceLabel: UILabel) {
```
mattbarker016: does this need to be a parameter, does it change? If so, maybe an enum or color or whatever distinguishes it

mattbarker016 (commented):
Can't wait to see this in action

TCAT/Assets.xcassets/live.imageset/Contents.json
myo3: Good point. Lmk when you add the other parameters and I can use that instead of the picture

TCAT/Cells/RouteTableViewCell.swift
myo3: yeah im not sure how i would combine them b/c each repeated ` self.setLiveElements(withStartTime: nil, withDelay: nil)` is for a different logic case (else, default, error for request, etc)

TCAT/Controllers/RouteOptionsViewController.swift
myo3: yeah that was for debugging. i'll delete that

myo3:
Implemented transfers and stopped distance label from running off as well as PR changes

TCAT/Cells/RouteTableViewCell.swift
mattbarker016: Maybe an enum for each scenario? I wanted to adjust the circle dot view, and realizes the three parameters were only being used in specific ways. It would maybe be a bit more understandable to have the cases be an enum

TCAT/Model/Direction.swift
```diff
@@ -67,7 +67,7 @@ class Direction: NSObject, NSCopying {
     var stops: [LocationObject] = []
     
     /// Whether the user should stay on this direction's bus for an upcoming transfer.
-    var stayOnBusTransfer: Bool = false
+    var stayOnBusForTransfer: Bool = false
```
mattbarker016, line 70: Good rename idea

mattbarker016 (commented):
Neato. Merge for now and I'll add my stuff and we can figure out the LiveIndicator and Circle views

TCAT/Cells/RouteTableViewCell.swift
myo3: Okay for me all I would need is a color parameter and size b/c looking at the Zeplin your RSS feed icon is 8.5 by 8.5 and mine's is 10.5 by 10.5. So we could have an enum like this:
```
enum RSSSize: Int {
    case small = 8.5
    case medium = 10.5
```

And then the init for the RSSFeedIcon could look like this:
```
init(size: RSSSize, color: UIColor) {
```

mattbarker016: I'll add that!

PR #93, opened 2018-03-08
matt/gen-updates

mattbarker016:
I roughly implemented a smaller **LiveIndicator**, but it will likely need to be tweaked, but the framework to do that is there. I also reworked the **Circle** view, which again needs to be looked at to be right across multiple screens. I think we should get this on master and just work together on the branch to iron out the issues. 

**Update:** Fixed basic errors from merge conflicts. I converted the LiveIndicator UIImageView to the LiveIndicator view. Hopefully color changes will work!


TCAT/AllStopsTableViewController.swift
```diff
@@ -126,9 +131,9 @@ class AllStopsTableViewController: UITableViewController {
         let section = sectionIndexes[sortedKeys[indexPath.section]]
         let optionsVC = RouteOptionsViewController()
         guard let busStopSelected = section?[indexPath.row]
-            else {
-                print("Could not find bus stop")
-                return
+        else {
```
myo3: prefer `else` on `guard let` line. Just reads better

TCAT/Cells/BusStopCell.swift
```diff
@@ -38,7 +38,7 @@ class BusStopCell: UITableViewCell {
         imageView?.frame = CGRect(x: 10.0, y: 5.0, width: imageWidth, height: imageHeight)
         imageView?.contentMode = .scaleAspectFit
         imageView?.center.y = bounds.height / 2.0
-        imageView?.image = reuseIdentifier == "currentLocation" ? #imageLiteral(resourceName: "location") : #imageLiteral(resourceName: "pin")
+        imageView?.image = reuseIdentifier == Key.Cells.currentLocationIdentifier ? #imageLiteral(resourceName: "location") : #imageLiteral(resourceName: "pin")
```
myo3: i think it's good you're replacing the string with a var  name. however, i think the cell identifier should not be in a global file but rather be a `let` variable at the top of the cell class b/c the only place you're going to use the identifier is in the cell.

TCAT/Cells/BusStopTableViewCell.swift
```diff
@@ -44,7 +44,7 @@ class BusStopTableViewCell: UITableViewCell {
         connectorBottom.backgroundColor = .tcatBlueColor
         contentView.addSubview(connectorBottom)
         
-        statusCircle = Circle(size: .small, color: .tcatBlueColor, style: .outline)
+        statusCircle = Circle(type: .smallOutline, color: .tcatBlueColor)
```
myo3: why the change? :( i don't see the point of adding this new enum that encapsulates the size and style b/c it makes the Circle class less flexible. Can we go back??

TCAT/Cells/RouteTableViewCell.swift
```diff
@@ -144,33 +143,33 @@ class RouteTableViewCell: UITableViewCell {
...
     func setLiveElements(withFirstDepartDirection firstDepartDirection: Direction?) {
+        
         if let direction = firstDepartDirection {
-            Network.getBusLocations(routeID: String(direction.routeNumber), tripID: direction.tripID, stopID: direction.startLocation.id)
-                .perform(withSuccess: { (result) in
-                    if let busLocation = result.busLocation {
+            
+            Network.getBusLocations([direction]).perform(withSuccess: { (result) in
+                
+                for busLocation in result.busLocations {
```
myo3: why is this a for loop? i only want the live update for the first depart direction so i would think i would only need `result.busLocations.first`. im just confused on why we're going through an array of `busLocation` when I really only want the delay for the bus that corresponds to the first depart direction.

TCAT/Cells/SearchResultsCell.swift
```diff
@@ -39,7 +39,7 @@ class SearchResultsCell: UITableViewCell {
         imageView?.contentMode = .scaleAspectFit
         imageView?.center.y = bounds.height / 2.0
         imageView?.image = #imageLiteral(resourceName: "pin")
-        imageView?.tintColor = reuseIdentifier == "cornellDestinations" ? .tcatBlueColor : .placeColor
+        imageView?.tintColor = reuseIdentifier == Key.Cells.cornellDestinationsIdentifier ? .tcatBlueColor : .mediumGrayColor
```
myo3: same comment. think resuseIdentifiers vars should be associated with the cell their in rather than a global Key struct

TCAT/Utilities/Network.swift
```diff
@@ -12,18 +12,24 @@ import CoreLocation
 import GooglePlaces
 
 class Network {
-
-    static let ipAddress = "54.174.47.32" // 10.132.9.134
-
+    
+    /// IP Address of server instance
+    static let ipAddress = "192.168.1.8" // Current Backend IP Address: 54.174.47.32
```
myo3: let's change that back to `54.174.47.32`

TCAT/Utilities/Network.swift
```diff
@@ -81,16 +87,30 @@ class Network {
...
-        request.parameters = [
-            "routeID"   :     routeID,
-            "tripID"    :     tripID,
-            "stopID"    :     stopID
-        ]
-        request.method = .get
+        
+        let departDirections = directions.filter { $0.type == .depart }
+
+        let dictionary = departDirections.map { (direction) -> [String : Any] in
```
myo3: nice use of swifty functional programming

TCAT/Utilities/Network.swift
```diff
@@ -161,15 +181,21 @@ class AllBusStops: JSONDecodable {
...
 
     required init(json: JSON) throws {
+        print("Bus Location JSON:", json)
         if json["success"].boolValue {
-            let data = json["data"]
-            busLocation = parseBusLocation(json: data)
+            if let data = json["data"].array {
+                busLocations = data.map { parseBusLocation(json: $0) }
+            } else {
+                busLocations = []
```
myo3: don't think this is necessary since have default value `[]` for `busLocations`

TCAT/Views/CircleLine.swift
```diff
@@ -42,11 +42,11 @@ class CircleLine: UIView {
     
     // MARK: Position
     
-    private func positionSolidCircle(){
+    private func positionSolidCircle() {
```
myo3: whoops ^^;

TCAT/Views/LiveIndicator.swift
```diff
@@ -24,39 +29,54 @@ class LiveIndicator: UIView {
     fileprivate let DIM_OPACITY: CGFloat = 0.5
     fileprivate let INTERVAL: TimeInterval = LiveIndicator.INTERVAL
     
+    /// The color to draw the indicator with
+    var color: UIColor = .white
```
myo3: not a huge fan of default values unless they properly describe a default way of initializing (like if you have an `init()`). Think its better if you initialize color before the super.init() calls of both inits, rather than provide a random default value.

TCAT/Views/LiveIndicator.swift
```diff
@@ -100,6 +120,11 @@ class LiveIndicator: UIView {
         }
     }
     
+    func setColor(to color: UIColor) {
+        self.color = color
+        drawViews()
```
myo3: wouldn't this add more layers and views on top of what's there? we should be actually changing the colors of the layer and views, not adding more

TCAT/Cells/BusStopCell.swift
mattbarker016: It's used in a few places

TCAT/Cells/BusStopTableViewCell.swift
mattbarker016: I'll revert, but different permutations of stuff are sorta untested

TCAT/Cells/RouteTableViewCell.swift
mattbarker016: Resolved with new `/delay` endpoint

TCAT/Cells/SearchResultsCell.swift
mattbarker016: It's used in multiple files though too

TCAT/Views/LiveIndicator.swift
mattbarker016: It's not random, it really is the default for the LiveIndicator. Do you think it should be a default parameter in the init instead?

TCAT/Views/LiveIndicator.swift
mattbarker016: Good call

TCAT/Views/LiveIndicator.swift
mattbarker016: I removed the default by the variable, left it in parameters

PR #101, opened 2018-03-14
Matt/map fixes

mattbarker016:
Whatever the latest I have here by 4 PM is getting put onto TestFlight for the Wednesday bug sesh.


TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -111,8 +114,25 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
...
         locationManager.requestWhenInUseAuthorization()
         locationManager.startUpdatingLocation()
+        
+        // Set up Share button
+//        let button = UIButton(frame: CGRect(x: 0, y: 0, width: 26, height: 38))
+//        button.setImage(#imageLiteral(resourceName: "share"), for: .normal)
+//        button.translatesAutoresizingMaskIntoConstraints = false
+//        button.widthAnchor.constraint(equalToConstant: 26).isActive = true
+//        button.heightAnchor.constraint(equalToConstant: 38).isActive = true
+//        let shareButton = UIBarButtonItem(customView: button)
```
myo3: prob would delete if don't need

TCAT/InformationViewController.swift
```diff
@@ -28,6 +28,10 @@ class InformationViewController: UIViewController {
         
         title = "About Us"
         
+        let scrollView = UIScrollView(frame: view.frame)
+        scrollView.contentSize = view.frame.size
+        view = scrollView
+
```
myo3: See what you're doing but think it would read better if you add `scrollView` to the view and then add all the other views using `scrollView.addSubview()`

myo3:
Reviewed PR. Looks good minus a couple of style changes

PR #116, opened 2018-04-01
Monica/usability bugs

myo3:
Bugs addressed:
- update current location on pull to refresh
- fixed bug where location services disabled alert showed up even though user enabled location services
- fixed bug where sometimes buses that are on time initially show up as late, but then change to on time about 5 sec later
- can now pull to refresh from empty state

### DIFF
--- a/TCAT/Controllers/HomeViewController.swift
+++ b/TCAT/Controllers/HomeViewController.swift
@@ -182,7 +182,7 @@ class HomeViewController: UIViewController, UITableViewDelegate, UITableViewData
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         locationManager.delegate = self
-        locationManager.requestAlwaysAuthorization()
+        locationManager.requestWhenInUseAuthorization()
     }
 
     func createSections() -> [Section] {
--- a/TCAT/Controllers/HomeViewController.swift
+++ b/TCAT/Controllers/HomeViewController.swift
@@ -182,7 +182,7 @@ class HomeViewController: UIViewController, UITableViewDelegate, UITableViewData
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         locationManager.delegate = self
-        locationManager.requestWhenInUseAuthorization()
+        locationManager.requestAlwaysAuthorization()
     }
 
     func createSections() -> [Section] {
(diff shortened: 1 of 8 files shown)


TCAT/Controllers/HomeViewController.swift
```diff
@@ -182,7 +182,7 @@ class HomeViewController: UIViewController, UITableViewDelegate, UITableViewData
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         locationManager.delegate = self
-        locationManager.requestWhenInUseAuthorization()
+        locationManager.requestAlwaysAuthorization()
```
mattbarker016: Why do we need always? Uber made headlines for only having this option, which basically lets developers potentially track you wherever. I'm very cautious about what apps I allow to do this. We should only have it if the app has live background navigation like Google Maps. We can refresh the location when the app comes into the foreground again through an AppDel function or something similar.

TCAT/Controllers/HomeViewController.swift
```diff
@@ -445,43 +445,41 @@ class HomeViewController: UIViewController, UITableViewDelegate, UITableViewData
...
-                    actions.append(UIAlertAction(title: "Dismiss", style: .default))
-                    leftButton = "Don't Remind Me Again"
-                } else {
-                    return
-                }
-            } else {
-                userDefaults.set(true, forKey: Constants.UserDefaults.locationAuthReminder)
+            let alertController = UIAlertController(title: "Location Services Disabled", message: "The app won't be able to use your current location without permission. Tap Settings to turn on Location Services.", preferredStyle: .alert)
+            let settingsAction = UIAlertAction(title: "Settings", style: .default) { (_) in
+                UIApplication.shared.open(URL(string: "App-prefs:root=LOCATION_SERVICES") ?? URL(string: UIApplicationOpenSettingsURLString)!, options: [:], completionHandler: nil)
```
mattbarker016, line 450: I didn't know you could do this!

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -402,51 +402,95 @@ class RouteOptionsViewController: UIViewController, UITableViewDelegate, UITable
...
 
     func locationManager(_ manager: CLLocationManager, didFailWithError error: Swift.Error) {
         print("RouteOptionVC locationManager didFailWithError: \(error.localizedDescription)")
-        locationManager.stopUpdatingLocation()
-        var locationAuthReminder = true
         
-        if let showReminder = userDefaults.value(forKey: Constants.UserDefaults.locationAuthReminder) as? Bool {
-            if showReminder {
-                locationAuthReminder = false
+        if error._code == CLError.denied.rawValue {
```
mattbarker016, line 414: Should we make this a global function in Extensions?

mattbarker016:
That's all!

TCAT/Controllers/HomeViewController.swift
myo3: Yeah i thought i needed for updating current location but just tested it and i don't so going to change back

TCAT/Controllers/HomeViewController.swift
myo3: yah i think it only works in iOS 11 and Swift 4 (haven't tested it for the older ones) so kept the other one cause that works no matter what

TCAT/Controllers/RouteOptionsViewController.swift
myo3: eh i think we only use it 2 places...let's keep it for now. But stmg I think we need to group up is the setting up the CLLocationManager. When I was doing this I realized we all set up our CLLocationManager slightly differently. I modified them so we all used the same attributes but we really should have 1 class or function that sets up the CLLocationManager uniformly

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -51,13 +51,13 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
     /** Initalize RouteDetailViewController. Be sure to send a valid route, otherwise
      * dummy data will be used. The directions parameter have logical assumptions,
      * such as ArriveDirection always comes after DepartDirection. */
-    init(route: Route, currentLocation: CLLocationCoordinate2D? = nil) {
```
mattbarker016, line 54: Are you sure you can get rid of this? I tried not having explicitly starting nil as a default and it complained

myo3: yeah just instead of `RouteDetailViewController(route: route)` you would do smtg like `RouteDetailViewController(route: route, currentLocation: nil)`. I thought I fixed that when I changed the initializer but maybe I missed a few

mattbarker016: So you're suggesting to not have default parameters when nil is involved?

myo3: Yeah cause that's the way other library functions do it. Like for completion handlers you just put in nil instead of having the default argument of a completion handler be nil.

mattbarker016: Ah, ok, I’ve done it on a few places. Will remember

PR #117, opened 2018-04-08
Matt/v1.0.1 fixes

mattbarker016:


mattbarker016:
If you're in Ithaca and if possible, please quickly run this on device to make sure nothing is freaking out. I will test when I get up there, but I'd appreciate another pair of eyes on the build running on the device.

TCAT/Controllers/RouteDetail+DrawerViewController.swift
```diff
@@ -230,62 +231,55 @@ class RouteDetailDrawerViewController: UIViewController, UITableViewDataSource,
...
-                }
-                
-                // Delay exists
+        let tripID = delayDirection.tripIdentifiers?.first
+        var stopID = delayDirection.stops.first?.id
+        
+        // Retrieve deleted stop ID (Route Parse) where first and last directions are removed.
+        if tripID != nil && stopID == nil {
+            stopID = route.rawDirections.first { (rawDir) -> Bool in
+                return rawDir.tripIdentifiers?.first == tripID
```
myo3: shouldn't you unwrap `tripID` in an `if-let` if going to use it?

TCAT/Controllers/RouteDetail+DrawerViewController.swift
```diff
@@ -230,62 +231,55 @@ class RouteDetailDrawerViewController: UIViewController, UITableViewDataSource,
...
-                    // Direction doesn't have a delay
+                    // print("Got delay of \(json["data"]["delay"].int ?? -1), reloading data")
+                    delayDirection.delay = json["data"]["delay"].int
+                    
+                    // Update delay variable of other ensuing directions
+                    
+                    self.directions.filter {
+                        let isAfter = self.directions.index(of: delayDirection)! < self.directions.index(of: $0)!
+                        return isAfter && $0.type != .depart
+                        }.forEach { (direction) in
```
myo3: put `.forEach` on new line/separate it from the `.filter`. Makes it easier to read

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -285,7 +285,11 @@ class RouteOptionsViewController: UIViewController, UITableViewDelegate, UITable
     // MARK: Process data
 
     func searchForRoutes() {
-        if let time = searchTime,
+      
+        if
+            let searchFrom = searchFrom,
```
myo3: put the `let` back next to the `if`

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -298,25 +302,74 @@ class RouteOptionsViewController: UIViewController, UITableViewDelegate, UITable
             // Prepare feedback on Network request
             let tapticGenerator = UIImpactFeedbackGenerator(style: .medium)
             tapticGenerator.prepare()
+            
+            let sameLocation = searchFrom.name == searchTo.name
```
myo3: noice naming of boolean values 👍

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -298,25 +302,74 @@ class RouteOptionsViewController: UIViewController, UITableViewDelegate, UITable
...
+            
+            /// Determine if coordinates in parameters are within range
+            func validCoordinates(_ requestParameters: [String : Any]?) -> Bool {
```
myo3: get rid of nested functions.  hard to read. 

Also why do you check validCoordinates from the `requestParameters` and not the `searchTo` and `searchFrom` values themselves? What edge case is this suppose to handle?

mattbarker016: Removed the nested functions and put on main level

We don't know the coordinates for google places until right before the request. It needs to be looked up via a Google API call. In that callback, we do the getRoutes network call, so once I have all the coordinates I check with the parameters generated. Given the current functions and classes that's the most accessible way to get the coords. I'd love to refactor this, I think it might sense to have one Place type and a variable called `type` that signifies whether it's a Google Place or Bus Stop. Then we can store coordinates once we look up the Google Place. Just one idea

TCAT/Controllers/RouteOptionsViewController.swift
mattbarker016: I kinda like the spacing it on the next line so they're all lined up. But it's your view so I'll put it back where I found it :P

TCAT/Controllers/RouteOptionsViewController.swift
myo3: Okay I think I can refactor it to be better.

Jw do you know any coordinates that are invalid? Surprised that Google maps doesn’t return valid coordinates all the time even tho it’s Google

PR #122, opened 2018-05-06
Matt/v1.0.2

mattbarker016:
**Monica's Changes:**

- Refactored `Network.swift` and `searchForRoutes()` in `RouteOptionsViewController.swift` to be more readable
- Created `JsonFileManager` to log route and delay jsons, search parameters (starting location name, ending location name, search time type, search time, etc), and request urls
- Disable 5 second looping delay request when move to `RouteDetailViewController` and enable when move back to `RouteOptionsViewController`


TCAT/Controllers/CustomNavigationController.swift
```diff
@@ -33,6 +34,25 @@ class CustomNavigationController: UINavigationController, UINavigationController
...
+                if let dot = viewControllerName.index(of: "."), let end = viewControllerName.index(of: ":") {
+                    let start = viewControllerName.index(dot, offsetBy: 1)
+                    location = String(viewControllerName[start..<end])
+                } else {
+                    location = viewControllerName
+                }
+                print("Screenshot Taken:", location)
+                let payload = ScreenshotTakenPayload(location: location)
+                RegisterSession.shared?.log(payload)
+            }
```
myo3: Logic is messy. Would use a `guard let` to get the `visibleViewController`. Also parsing the default string is very messy. So I looked into a couple of alternatives:
```
        print("HomeVC: debugDescription: \(debugDescription)")
        print("HomeVC: description: \(description)")
        print("HomeVC class: \(self.classForCoder)")
        print("HomeVC string describing: \(String(describing: HomeViewController.self))")
        print("HomeVC typeOf: \(type(of: self))")
```
Which printed like so:
```
HomeVC: debugDescription: <TCAT.HomeViewController: 0x7f9df5889c00>
HomeVC: description: <TCAT.HomeViewController: 0x7f9df5889c00>
HomeVC class: HomeViewController
HomeVC string describing: HomeViewController
HomeVC typeOf: HomeViewController
```
So I'm thinking its better to use one of the 3 options instead of `description` or `debugDescription` and parse the string. I especially think `type(of:)` might help you since it returns the dynamic type which may be helpful for you since you don't know the type of your VC but you want to get its name

TCAT/Controllers/InformationViewController.swift
```diff
@@ -184,16 +185,111 @@ class InformationViewController: UIViewController, UITableViewDataSource, UITabl
...
+        if MFMailComposeViewController.canSendMail() {
+        
+            let mailComposerVC = MFMailComposeViewController()
+            mailComposerVC.mailComposeDelegate = self
+            var didRetrieveLog = false
+
+            if let logs = retrieveLogs() {
+                mailComposerVC.addAttachmentData(logs, mimeType: "application/zip", fileName: "Logs.zip")
+                didRetrieveLog = true
+            }
```
myo3: Also one thing i forgot to mention, can u delete the zip after you send the email? I have this function `deleteZip()` in `JsonFileManager` that does that. I think it'd be good to delete the zip so that we don't take up too much space on the user's device

TCAT/Controllers/RouteDetail+DrawerViewController.swift
```diff
@@ -298,6 +272,53 @@ class RouteDetailDrawerViewController: UIViewController, UITableViewDataSource,
...
     }
+    
+    func tableView(_ tableView: UITableView, viewForFooterInSection section: Int) -> UIView? {
+        
+        let latitude = route.endCoords.latitude
+        let longitude = route.endCoords.longitude
+        let message = LocationPhrases.generateMessage(latitude: latitude, longitude: longitude)
+        let usePhraseFooterView = message != nil
+        
+        if usePhraseFooterView, let message = message {
```
myo3: change to `if let message = message`

myo3: Also might consider using a `guard let` with `message` and then in the `else` body return the empty footer view

TCAT/Controllers/RouteDetail+DrawerViewController.swift
```diff
@@ -298,6 +272,53 @@ class RouteDetailDrawerViewController: UIViewController, UITableViewDataSource,
...
+                screenBottom -= view.safeAreaInsets.bottom
+            }
+            
+            // Calculate height of space between last cell and the bottom of the screen, also accounting for summary
+            var footerHeight = screenBottom - (tableView.cellForRow(at: lastCellIndexPath)?.frame.maxY ?? screenBottom) - summaryView.frame.height
+            footerHeight = expandedCell != nil ? 0 : footerHeight
+            
+            emptyFooterView.frame.size = CGSize(width: view.frame.width, height: footerHeight)
+            emptyFooterView.contentView.backgroundColor = .white
+            emptyFooterView.layoutIfNeeded()
```
myo3, line 309: jw why do you have `emptyFooterView.layoutIfNeeded()`? also why do you add a tap gesture recognizer to you footer? jw where this footer is in your tableview

TCAT/Utilities/Phrases.swift
```diff
@@ -0,0 +1,245 @@
...
+    func isWithinRange(latitude: Double, longitude: Double) -> Bool {
+        let isLatInRange = minimumLatitude <= latitude && latitude <= maximumLatitude
+        let isLongInRange = minimumLongitude <= longitude && longitude <= maximumLongitude
+        return isLatInRange && isLongInRange
+    }
+    
+}
+
+/// Select random string from array
+func selectMessage(from messages: [String]) -> String {
```
myo3: not a fan of global functions. put that in a class

TCAT/Utilities/Phrases.swift
```diff
@@ -0,0 +1,245 @@
...
+//  Personality.swift
+//  Brella, I mean, TCAT
+//
+//  Created by Matthew Barker on 04/15/18
+//  Copyright © 2018 cuappdev. All rights reserved.
+//
+
+import Foundation
+
+struct Messages {
```
myo3, line 11: like the struct. Organization on point 👌

TCAT/Utilities/Phrases.swift
```diff
@@ -0,0 +1,245 @@
...
+    
+}
+
+
+class LocationPhrases {
+    
+    // For new places, use: https://boundingbox.klokantech.com set to CSV.
+    
+    // For overlapping places, put the smaller one first
+    static let places: [CustomLocation] = [
```
myo3, line 116: oohh more organization! :D

TCAT/Utilities/Phrases.swift
myo3: In fact, I would create a class `Phrases` that has `selectMessage(from:)` and then have `WalkingPhrases` and `LocationPhrases` subclass from `Phrases`. That way they can both access `selectMessage(from:)` and `selectMessage(from:)` does not have to be global

TCAT/Utilities/StoreKitHelper.swift
```diff
@@ -0,0 +1,71 @@
...
+        appOpenCount += 1
+        userDefaults.set(appOpenCount, forKey: APP_OPENED_COUNT)
+    }
+    
+    /// Ask for review at appropriate times in app lifecycle (not guaranteed to fire, Apple-controlled logic)
+    /// - override: Force a request of a review, overriding number of app launches
+    static func checkAndAskForReview(override: Bool = false) {
+        
+        if override {
+            StoreReviewHelper().requestReview()
```
myo3: this is poor style. better to have a shared instance of `StoreReviewHelper` to call instance methods on it. Look at `JSONFileManager` for reference. Basically create a singleton within your class

TCAT/Utilities/StoreKitHelper.swift
```diff
@@ -0,0 +1,71 @@
...
+//  Created by Matthew Barker on 04/15/18
+//  Copyright © 2018 cuappdev. All rights reserved.
+//
+//  https://gist.github.com/abhimuralidharan/fc717fb27d1d7388524e70a09860a786#file-storereviewhelper-swift
+//
+
+import UIKit
+import StoreKit
+
+struct StoreReviewHelper {
```
myo3: make class

TCAT/Views/PhraseLabelFooterView.swift
```diff
@@ -0,0 +1,51 @@
...
+            label.text = labelText
+            contentView.addSubview(label)
+        }
+
+        label.snp.makeConstraints { (make) in
+            let constant: CGFloat = 8
+            make.leading.equalToSuperview().offset(RouteDetailCellSize.regularWidth)
+            make.trailing.equalToSuperview().offset(-constant)
+            // make.top.equalToSuperview().offset(4)
+            // make.bottom.equalToSuperview().offset(4)
```
myo3: delete if not needed

TCAT/Views/PhraseLabelFooterView.swift
```diff
@@ -0,0 +1,51 @@
...
+    var label: UILabel = {
+        let label = UILabel()
+        label.text = ""
+        label.font = UIFont(name: Constants.Fonts.SanFrancisco.Regular, size: 12)
+        label.textColor = .lightGray
+        label.numberOfLines = 0
+        label.textAlignment = .left
+        label.lineBreakMode = .byWordWrapping
+        return label
+    }()
```
myo3: better to move this styling and initialization of label into `init(reuseIdentifer:)`. It better tells the story of initialization

TCAT/Views/PhraseLabelFooterView.swift
```diff
@@ -0,0 +1,51 @@
...
+//
+
+import UIKit
+import SnapKit
+
+class PhraseLabelFooterView: UITableViewHeaderFooterView {
+
+    var label: UILabel = {
+        let label = UILabel()
+        label.text = ""
```
myo3: unnecessary. delete

TCAT/Views/SummaryView.swift
```diff
@@ -11,11 +11,7 @@ import UIKit
...
     /// The route being used for the summary view
-    var route: Route! {
-        didSet {
-            setRoute()
-        }
-    }
+    var route: Route!
     
     /// The puller tab used to indicate dragability
     fileprivate var tab = UIView(frame: CGRect(x: 0, y: 6, width: 32, height: 4))
```
myo3: get rid of alllll `fileprivate` and make `private`. `fileprivate` not needed

TCAT/Views/SummaryView.swift
```diff
@@ -128,7 +124,9 @@ class SummaryView: UIView {
             
             mainLabel.attributedText = attributedString
             
+            // Find time within label to place live indicator
             if let stringRect = mainLabel.boundingRect(of: departDirection.startTimeWithDelayDescription + " ") {
+                print("setting origin")
```
myo3: is print statement needed? seemed that ur using it for debugging. if so, delete.

TCAT/Views/SummaryView.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  LiveIndicator.swift
+//  SummaryView.swift
```
myo3, line 2: Also just a general comment: If im not mistaken, this is the view you use to show the summary card with the bus icon and live icon (like the top view in the tableview in your viewcontroller)? Cause for that i would suggest you move to constraints. I imagine you do a lot of pixel point manipulation that's very manual and could be simplified with constraints

myo3 (dismissed):
Like the Phrases class—looks organized. Added some styling comments and suggested edits.

TCAT/Utilities/Phrases.swift
mattbarker016: I was trying to have the shared `messages` variable and shared` selectMessage` in a `Phrases` class and then subclass, but I was having issues with static functions and variables. Looks like circling back to the initial idea based on my solution works well

TCAT/Controllers/RouteDetail+DrawerViewController.swift
mattbarker016: Think I was trying to use the variable as a description for the logic. I don't like `guard let` here, we go from defining a message to handling an error case, back to using the message. Better to group similar stuff closer together

TCAT/Controllers/CustomNavigationController.swift
mattbarker016: Thanks!

PR #136, opened 2018-08-27
matt/swiftlint-init

### LINKED ISSUE #126: Install SwiftLint
Install library and then assign members to fix warnings.

mattbarker016:
Resolves #126 so we can delegate fixing the oodles of warnings

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -36,7 +36,6 @@ target 'TCAT' do
 
   # Other
   pod 'Fuzzywuzzy_swift', :git=> 'https://github.com/lxian/Fuzzywuzzy_swift.git'
-  pod 'SwiftLint'
 
   target 'TCATTests' do
     inherit! :search_paths
--- a/Podfile
+++ b/Podfile
@@ -36,6 +36,7 @@ target 'TCAT' do
 
   # Other
   pod 'Fuzzywuzzy_swift', :git=> 'https://github.com/lxian/Fuzzywuzzy_swift.git'
+  pod 'SwiftLint'
 
   target 'TCATTests' do
     inherit! :search_paths
--- a/TCAT/Cells/LargeDetailTableViewCell.swift
+++ b/TCAT/Cells/LargeDetailTableViewCell.swift
@@ -9,22 +9,22 @@
 import UIKit
 
 class LargeDetailTableViewCell: UITableViewCell {
-    
+
     var iconView: DetailIconView!
     var titleLabel: UILabel!
     var detailLabel: UILabel!
     var busIconView: BusIcon!
     var chevron: UIImageView!
-    
+
     let paragraphStyle = NSMutableParagraphStyle()
     let cellWidth: CGFloat = RouteDetailCellSize.regularWidth
     var cellHeight: CGFloat = RouteDetailCellSize.largeHeight
-    
+
     var direction: Direction!
     var isExpanded: Bool = false
     let edgeSpacing: CGFloat = 16
     let labelSpacing: CGFloat = 4
-    
+
     func getChevron() -> UIImageView {
         let chevron = UIImageView()
         chevron.frame.size = CGSize(width: 13.5, height: 8)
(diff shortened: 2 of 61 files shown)


Podfile
```diff
@@ -36,6 +36,7 @@ target 'TCAT' do
 
   # Other
   pod 'Fuzzywuzzy_swift', :git=> 'https://github.com/lxian/Fuzzywuzzy_swift.git'
+  pod 'SwiftLint'
```
AAAstorga: why do you believe a pod is better?

AAAstorga (commented):
Love all the swift autocorrect. Just need to talk about why use a pod. Just clutters things. No real benefit. You need the brew install to be able to run swiftlint autocorrect I believe

AAAstorga (commented):
Love all the swift autocorrect. Just need to talk about why use a pod. Just clutters things. No real benefit. You need the brew install to be able to run swiftlint autocorrect I believe

AAAstorga (commented):
lets have a discussion about why you want to use the SwiftLint pod vs just brew

TCAT/Cells/LargeDetailTableViewCell.swift
```diff
@@ -55,77 +55,77 @@ class LargeDetailTableViewCell: UITableViewCell {
...
+
     required init?(coder aDecoder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
-    
+
     override func prepareForReuse() {
-        
+
 //        busIconView.prepareForReuse()
```
myo3, line 79: delete commented out code if unnecessary

TCAT/Cells/LargeDetailTableViewCell.swift
```diff
@@ -55,77 +55,77 @@ class LargeDetailTableViewCell: UITableViewCell {
...
     func formatTitleLabel(_ label: UILabel) -> UILabel {
-        
+
         // Set inital text in label
         label.text = direction.type == .transfer ? "Bus becomes" : "Board"
- 
+
         // Add correct amount of spacing to create a gap for the busIcon
         // Using constant always returned from
         //      while label.frame.maxX < busIconView.frame.maxX + 8 {
```
myo3, line 131: delete commented out code

TCAT/Cells/LargeDetailTableViewCell.swift
```diff
@@ -136,61 +136,60 @@ class LargeDetailTableViewCell: UITableViewCell {
...
         let content = label.text! + direction.locationNameDescription
         let attributedString = bold(pattern: direction.name, in: content)
         label.attributedText = attributedString
         paragraphStyle.lineSpacing = 4
-        
+
         label.numberOfLines = 0
         label.sizeToFit()
         label.frame.size.width = (chevron.frame.minX - 12) - cellWidth
         label.frame.origin.y = edgeSpacing // - paragraphStyle.lineSpacing
```
myo3, line 150: delete commented out code

TCAT/Cells/LargeDetailTableViewCell.swift
```diff
@@ -136,61 +136,60 @@ class LargeDetailTableViewCell: UITableViewCell {
...
     /** Abstracted formatting of content for detailLabel. Needs titleLabel */
     func formatDetailLabel(_ label: UILabel, _ titleLabel: UILabel) -> UILabel {
-        
+
         label.text = "\(direction.stops.count + 1) stop\(direction.stops.count + 1 == 1 ? "" : "s")"
-        
+
         // Number of minutes for the bus direction
         var timeString = Time.timeString(from: direction.startTime, to: direction.endTime)
         if timeString == "0 min" {  timeString = "1 min" }
```
myo3, line 167: better to check on an int rather than string in case we change the format of our strings. i think we should have a method in our `Time` class that does that but could be wrong

myo3 (approved):
some miscellaneous comments, but main change looks good

TCAT/Cells/LargeDetailTableViewCell.swift
mattbarker016: This is outside the scope of this PR. This is only meant to generate warnings so we can then refactor and fix code. This will be fixed in a future PR.

Issue #152, opened 2018-09-24
Standardize and Update Font Usage


mattbarker016:
A quick UI fix for the next updated version of the app.

## Instructions

1. We currently use `UIFont.systemFont`, `UIFont.boldSystemFont`, and `Constants.Fonts.SanFrancisco` (e.g. "SFUIText-xxx") throughout the app. Replace uses of `UIFont.systemFont | UIFont.boldSystemFont` with an initializer using the `Constants.Fonts.SanFrancisco` enum. 

- [x] Consolidate San Francisco and System Font usage to one font.

2. Download the San Francisco font [here](https://developer.apple.com/fonts/). We are using the following variants of **SF-Pro-Display**
- Regular
- Medium
- Bold
- Semibold

- [x] Install Pro Display fonts

3. Remove previously included fonts and install and include them following [this guide](https://developer.apple.com/documentation/uikit/text_display_and_fonts/adding_a_custom_font_to_your_app). You'll need to change the string identifiers in `Constants.swift`.

- [x] Properly add new fonts

4. Use one consistent font initializer everywhere within the app.

- [x] Go through app and use one system for fonts.

mattbarker016:
**Update:** Just use SF-Pro-Display everywhere. Updated above issue

mattbarker016:
To be fully completed in upcoming Styling PR

PR #164, opened 2018-10-13
#152 - Update and standardize font usage

mattbarker016:
#152 

This is a **phat** PR. Most changes are the same syntax updates to instantiating UIFont. The most important changes are in `Styles.swift` and `Extensions.swift`. I would look close there first and skim the rest if interested.

**Extensions.swift** 

Changes to the bold(...) function. I added additional parameters to make the bolding process less reliant on .boldSystemFont and the like. It's a bit clunky, but once we trim down fonts from the design audit, we can re-assess this.

**Styles.swift**

Eventual home of all colors, fonts, and spacing for the entire app! Set the boilerplate up, but only focusing on fonts for now.


TCAT/Utilities/Extensions.swift
```diff
@@ -372,28 +372,24 @@ func presentShareSheet(from view: UIView, for route: Route, with image: UIImage?
...
     
-    let boldFontAttribute = [NSAttributedString.Key.font : UIFont.boldSystemFont(ofSize: fontSize)]
+    let boldFontAttribute = [NSAttributedString.Key.font : boldFont]
     
     do {
         let regex = try NSRegularExpression(pattern: pattern, options: [])
-        let ranges = regex.matches(in: string, options: [], range: NSMakeRange(0, string.count)).map { $0.range }
+        let ranges = regex.matches(in: plain_string, options: [], range: NSMakeRange(0, plain_string.count)).map { $0.range }
         for range in ranges { newAttributedString.addAttributes(boldFontAttribute, range: range) }
     } catch { }
```
myo3: catch the particular error and print an error statement

TCAT/Utilities/Extensions.swift
```diff
@@ -372,28 +372,24 @@ func presentShareSheet(from view: UIView, for route: Route, with image: UIImage?
 /** Bold a phrase that appears in a string, and return the attributed string.
     Only shows the last bolded phrase.
  */
-func bold(pattern: String, in string: String) -> NSMutableAttributedString {
-    let fontSize = UIFont.systemFontSize
-    let fontAttribute = [NSAttributedString.Key.font : UIFont.systemFont(ofSize: fontSize)]
+func bold(pattern: String, in string: String, from font: UIFont, to boldFont: UIFont) -> NSMutableAttributedString {
```
myo3: it'd be nice if this could be an extension in one of the string classes. that way its not a global func

TCAT/Utilities/Extensions.swift
```diff
@@ -372,28 +372,24 @@ func presentShareSheet(from view: UIView, for route: Route, with image: UIImage?
...
+    let fontAttribute = [NSAttributedString.Key.font : font]
     let attributedString = NSMutableAttributedString(string: string, attributes: fontAttribute)
-    return bold(pattern: pattern, in: attributedString)
+    return bold(pattern: pattern, in: attributedString, to: boldFont)
 }
 
 /** Bold a phrase that appears in an attributed string, and return the attributed string */
-func bold(pattern: String, in attributedString: NSMutableAttributedString) -> NSMutableAttributedString {
+func bold(pattern: String, in string: NSMutableAttributedString, to boldFont: UIFont) -> NSMutableAttributedString {
```
myo3: private this method if only using it for the previous bold func

TCAT/Utilities/Extensions.swift
myo3: also for args i think having longer args names helps explain what the args are for. 

So:
- `pattern` -> `textToBeBold`
- `in` -> `containerText`
- `from` -> `originalFont`
-  `boldFont` think is good

myo3 (approved):
Some stylistic comments

TCAT/Utilities/Extensions.swift
mattbarker016: I removed it because it never fires. Added back

TCAT/Utilities/Extensions.swift
mattbarker016: Both are needed, but made it an extension

TCAT/Cells/BusStopCell.swift
```diff
@@ -44,7 +44,7 @@ class BusStopCell: UITableViewCell {
 
         textLabel?.frame = CGRect(x: labelXPosition, y: 0.0, width: frame.width - labelWidth, height: labelHeight)
         textLabel?.center.y = bounds.height / 2.0
-        textLabel?.font = .systemFont(ofSize: 13)
+        textLabel?.font = UIFont.style(Fonts.System.regular, size: 13)
```
AAAstorga: thoughts on dropping `UIFont` and doing just `.style(...)`

TCAT/Utilities/Styles.swift
```diff
@@ -0,0 +1,55 @@
...
+
+// MARK: - Fonts
+
+/// Font identifiers
+struct Fonts {
+    
+    struct SanFrancisco {
+        
+        // Pro Display (New)
+//        static let regular = "SFProDisplay-Regular"
```
AAAstorga: delete if we won't use these

AAAstorga (approved):
LGTM, just 2 small comments

TCAT/Utilities/Styles.swift
mattbarker016: We *may* use it... not sure yet. Should I delete and store this locally until we know, or can it just stay in here for now? It won't stay like this for long

TCAT/Cells/BusStopCell.swift
mattbarker016: Didn't know you could do that for a function! (not sure why, it makes sense)

Issue #134, opened 2018-08-27
Update models to have codable support


AAAstorga:
Makes things a lot cleaner in the long term. Should also write a bible page on codable.

PR #184, opened 2018-11-18
Codable

Omarrasheed:
#134


TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -706,7 +704,7 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
 
     /// Create fake bus for debugging and testing bus indicators
     func createDebugBusIcon() {
```
myo3, line 709: why is this method here? is it used in the app?

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -407,37 +406,30 @@ class RouteOptionsViewController: UIViewController, DestinationDelegate, SearchB
...
 
-        request.perform(withSuccess: { routeJson in
-                            self.processRouteJson(routeJSON: routeJson, requestUrl: requestUrl)
-                        },
-                        failure: { requestError in
-                            self.processRequestError(error: requestError, requestUrl: requestUrl)
-                        }
-        )
+        request.performCollectingTimeline { (response) in
+            do { try print(JSON.init(data: response.data!)) } catch { print("error") }
```
myo3: are u printing the json for debugging? if so get rid of printing bef merge to master

myo3: Also would recommend printing out a more descriptive error statement than "error". I usually like printing the class and method i'm in and then describing the error (e.g. "No response data")

E.g. printing an error w file name and function name in Swift:
```
let fileName = "RouteOptionsViewController"
let line = "\(fileName) \(#function): \(error)"
print(line)
```

myo3: Also along the lines of printing an error. it would be easier to know what error-ed out if you use an if-let for response.data. B/c there's 2 things i can see that could go wrong 1) response.data is nil or 2) try statement fails. In both cases you should print different errors

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -407,37 +406,30 @@ class RouteOptionsViewController: UIViewController, DestinationDelegate, SearchB
...
+            switch response.result {
+            case .success(let routesResponse):
+                if let data = response.data { do { try JSONFileManager.shared.saveJSON(JSON.init(data: data), type: .routeJSON) } catch { print("error") } }
```
myo3: same comment about more descriptive error printing

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+    
+    func getServiceAlerts() {
+        Network.getAlerts().perform(withSuccess: { (request) in
+            if (request.success) {
+                for alert in request.data {
+                    self.alerts.append(alert)
+                }
+            }
+        }) { (error) in
+            print(error)
```
myo3: same comment about printing errors (e.g. print file name, function, etc)

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+    
+    func filterAlertPriorities() {
+        
+    }
+}
+
+extension ServiceAlertsViewController: DZNEmptyDataSetSource {
```
myo3: assuming these will get filled? else delete these extensions

TCAT/Model/Direction.swift
```diff
@@ -36,47 +36,90 @@ class Direction: NSObject, NSCopying {
      - depart: The name of the bus stop where the bus is departing from
      - arrive: The name of the bus stop where the user gets off the bus
      */
-    var name: String
+    var name: String {
+        didSet {
+            print("Old Value: \(oldValue)")
```
myo3: print class name and function (in this case its a property). E.g. "Direction's Name Old Value"

TCAT/Model/Direction.swift
```diff
@@ -36,47 +36,90 @@ class Direction: NSObject, NSCopying {
...
+    required init (from decoder: Decoder) {
+        let container = try! decoder.container(keyedBy: CodingKeys.self)
+        type = try! container.decode(DirectionType.self, forKey: .type)
+        name = try! container.decode(String.self, forKey: .name)
+        startLocation = try! container.decode(LocationObject.self, forKey: .startLocation)
+        endLocation = try! container.decode(LocationObject.self, forKey: .endLocation)
+        startTime = Date.parseDate(try! container.decode(String.self, forKey: .startTime))
+        endTime = Date.parseDate(try! container.decode(String.self, forKey: .endTime))
+        path = try! container.decode([CLLocationCoordinate2D].self, forKey: .path)
+        do { routeNumber = try container.decode(Int.self, forKey: .routeNumber) } catch { routeNumber = 0 }
```
myo3: dont think routeNumber should default to 0. If we don't have a route number it should be an error (unless some directions don't have route numbers). at the very least routeNumber should default to -1 b/c there's no way for a routeNumber to be -1. tbh would make routeNumber optional so u can have a nil state to represent no routeNumber

TCAT/Model/LocationObject.swift
```diff
@@ -10,47 +10,61 @@ import UIKit
...
+        case longitude = "long"
+        case name
+        case id = "stopID"
+    }
+
+    required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        latitude = try container.decode(Double.self, forKey: .latitude)
+        longitude = try container.decode(Double.self, forKey: .longitude)
+        do {name = try container.decode(String.self, forKey: .name)} catch {name = ""}
```
myo3: dont feel like should have default value of empty string for name. if LocationObject has a name it should be its name and if LocationObject doesn't then name should be nil or an error if all LocationObject should have a name

TCAT/Model/LocationObject.swift
```diff
@@ -10,47 +10,61 @@ import UIKit
...
+        do {id = try container.decode(String.self, forKey: .id)} catch {id = ""}
```
myo3: same comment for id. If all LocationObjects should have an id this should error out. if not LocationObject id property should be optional

TCAT/Utilities/Extensions.swift
```diff
@@ -310,9 +265,46 @@ extension CLLocationCoordinate2D {
...
+        dateFormatter.locale = Locale(identifier: "en_US_POSIX")
+        dateFormatter.dateFormat = "yyyy-MM-dd'T'HH:mm:ssZZZZ"
+        return dateFormatter
+    }()
+}
+
+extension Date {
+    static func parseDate(_ dateString: String) -> Date {
+        let dateFormatter = DateFormatter.defaultParser
+        let date = dateFormatter.date(from: dateString) ?? Date.distantPast
```
myo3, line 298: why is the default `Date.distantPast`? it would be an error if `dateFormatter.date(from: dateString)` b/c u want to get the `Date` from the `dateString` and not a random default date. but also dont know how this is used in implementation, but also still feel that using default values (unless they make sense) is improper style.

TCAT/Utilities/Network+Models.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        success = try container.decode(Bool.self, forKey: .success)
+        data = try container.decode([BusStop].self, forKey: .data)
+        parseAllStops()
+    }
+
+    func parseAllStops() {
```
myo3, line 70: write comment to explain what parseAllStops does. esp for lines 81-97

myo3 (approved):
left some stylistic and model comments. glad to see things are moving to codable!

TCAT/Controllers/RouteOptionsViewController.swift
Omarrasheed: Oh whoops that was for debugging lol I'll take that out

TCAT/Controllers/RouteDetail+ContentViewController.swift
Omarrasheed: Not really sure.. I don't think I had made this method? @mattbarker016

TCAT/AppDelegate.swift
```diff
@@ -123,9 +123,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     /* Get all bus stops and store in userDefaults */
     func getBusStops() {
-        Network.getAllStops().perform(withSuccess: { stops in
-            let allBusStops = stops.allStops
-            if allBusStops.isEmpty {
+        Network.getAllStops().perform(withSuccess: { AllStopsRequest in
```
mattbarker016: nitpick: looks like a type, should be lowercase

TCAT/Cells/RouteTableViewCell.swift
```diff
@@ -179,247 +178,237 @@ class RouteTableViewCell: UITableViewCell {
...
                         self.setDepartureTimeAndLiveElements(withRoute: route)
                         return
                     }
-                    
                     let isNewDelayValue = (route.getFirstDepartRawDirection()?.delay != delay)
                     if isNewDelayValue {
                         JSONFileManager.shared.logDelayParemeters(timestamp: Date(), stopId: stopId, tripId: tripId)
                         JSONFileManager.shared.logURL(timestamp: Date(), urlName: "Delay requestUrl", url: Network.getDelayUrl(tripId: tripId, stopId: stopId))
-                        JSONFileManager.shared.saveJSON(json, type: .delayJSON(rowNum: self.rowNum ?? -1))
+                        // JSONFileManager.shared.saveJSON(json, type: .delayJSON(rowNum: self.rowNum ?? -1))
```
mattbarker016: Uncomment this

TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -239,8 +239,8 @@ extension AllStopsTableViewController: DZNEmptyDataSetSource, DZNEmptyDataSetDel
     }
 
     func retryNetwork(completion: @escaping () -> Void) {
-        Network.getAllStops().perform(withSuccess: { stops in
-            let allBusStops = stops.allStops
+        Network.getAllStops().perform(withSuccess: { AllStopsRequest in
```
mattbarker016: Same nitpick

TCAT/Controllers/RouteDetail+ContentViewController.swift
mattbarker016: Does exactly what the comment says

TCAT/Controllers/RouteDetail+DrawerViewController.swift
```diff
@@ -144,41 +144,41 @@ class RouteDetailDrawerViewController: UIViewController, UITableViewDataSource,
...
+
     func drawerPositionDidChange(drawer: PulleyViewController, bottomSafeArea: CGFloat) {
-        
+
         justLoaded = false
-        
+
         // Center map on drawer change
-        if drawer.drawerPosition == .collapsed || drawer.drawerPosition == .partiallyRevealed  {
+        if drawer.drawerPosition == .collapsed || drawer.drawerPosition == .partiallyRevealed {
```
mattbarker016: nice

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,90 @@
...
+
+        tableView = UITableView(frame: .zero, style: .grouped)
+        tableView.backgroundColor = view.backgroundColor
+        tableView.delegate = self
+        tableView.dataSource = self
+        tableView.emptyDataSetSource = self
+        tableView.emptyDataSetDelegate = self
+        
+        tableView.separatorColor = .lineDotColor
+        tableView.keyboardDismissMode = .onDrag
```
mattbarker016: Is there ever a keyboard on this screen? Why do you need this?

TCAT/Model/BusLocation.swift
```diff
@@ -32,16 +32,42 @@ class BusLocation: NSObject {
...
     private var _iconView: UIView?
-    
+
+    private let noDataString = "noData"
+    private let validDataString = "validData"
+    private let invalidDataString = "invalidData"
+
+    private enum CodingKeys: String, CodingKey {
+        case dataType = "case"
+        case destination
```
mattbarker016, line 47: Just confirming, do you need to list all of these if you only have to change one CodingKey?

TCAT/Model/Direction.swift
mattbarker016: Why are you printing here?

TCAT/Model/Direction.swift
mattbarker016: Off of this discussion, remember how we say bus routes that said "13X"? Might be better to do this as a string... I'll see backend's thoughts too

mattbarker016: Having it be 0 was useful for debugging when some routes ended up going wrong, but maybe an empty bus icon could be better (e.g. better to say no route number than a wrong number route)

mattbarker016: Point is, we should handle this error gracefully so that we present everything else we can

TCAT/Model/LocationObject.swift
mattbarker016: Again, while it should fail, this failing should minimally impact everything else presented in the view

mattbarker016 (commented):
Meaty! Take a look at our comments, message us for any clarification / confusion. Great stuff!

mattbarker016:
Don't forget about the merge conflicts too!

TCAT/Model/Direction.swift
myo3: yeah i agree--an empty bus icon would be better than "route 0". And i see you point about not crashing the app even if we get bus routes that say "13X". I still think we can handle it by making routeNumber optional and then if it is nil when creating a bus icon, create an empty bus icon instead

TCAT/Model/BusLocation.swift
Omarrasheed: Yea unfortunately

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -183,10 +183,11 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
 
         // Live Tracking Network Timer
         liveTrackingNetworkTimer?.invalidate()
-        liveTrackingNetworkTimer = Timer.scheduledTimer(timeInterval: liveTrackingNetworkRefreshRate, target: self, selector: #selector(getBusLocations),
-                                            userInfo: nil, repeats: true)
-        liveTrackingNetworkTimer?.fire()
-
+        if directions.contains(where: { $0.type != .walk }) {
```
mattbarker016, line 186: Were we calling our backend for walking directions? Good move to stop the call here as opposed to further down the line.

TCAT/Model/Route.swift
```diff
@@ -159,7 +159,7 @@ class Route: NSObject, Codable {
         for (index, direction) in rawDirections.enumerated() {
             if index != last && index != first && direction.type == .walk {
                 direction.type = .arrive
-                direction.name = rawDirections[index - 1].endLocation.name
+                direction.name = rawDirections[index - 1].stops.last?.name ?? "Nil"
```
mattbarker016: nitpick: While this should never appear, I think "Bus Stop" would look be a bit better of an error state than "Nil" to show users

TCAT/Model/Route.swift
```diff
@@ -296,12 +296,12 @@ class Route: NSObject, Codable {
             noDepartDirection = false
 
             let number = direction.routeNumber
-            let start = direction.startLocation.name
-            let end = direction.endLocation.name
-            var line = "take Route \(number) from \(start) to \(end). "
+            let start = direction.stops.first?.name
```
mattbarker016: nitpick: The string will look bad if the optional `start` and `end` fail. Add a `??` and call them something like "the starting location" and "the ending location". That way, the sentence stays readable and just looks redundant, and not incorrect 🙃

mattbarker016 (commented):
Just some tiny nitpicks. LGTM after you fix them, unless you have any questions you can merge in. Just wait for @myo3 to check this out too!

mattbarker016:
This looks good to merge in. I can't find the comment, but did you fix that one issue where you commented a write to JSON line that I mentioned to uncomment? Merge it in otherwise!

PR #192, opened 2019-01-28
Analytics + Debug Server

### LINKED ISSUE #182: Created routeSelected endpoint
We are going to log which routes users select on the client side. We will pass along the index of the route to backend. For now, this will be a single index, but once we start sorting routes into sections, it will include a `sectionIndex` and a `rowIndex`. This should be done on `didSelect`.

mattbarker016:
Added /routeSelected endpoint to report to backend what routes are selected, using `routeId`. I'm waiting for backend for confirmation on the specs, but I figured I'd queue this up.

Resolves #182.

### DIFF
--- a/TCAT/Model/Route.swift
+++ b/TCAT/Model/Route.swift
@@ -51,6 +51,9 @@ class Route: NSObject, Codable {
 
     /// The time a user arrives at their destination.
     var arrivalTime: Date
+    
+    /// A unique identifier for the route
+    var routeId: String
 
     /// The amount of time from now until the departure
     var timeUntilDeparture: DateComponents {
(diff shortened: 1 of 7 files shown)


mattbarker016:
Completes #193

TCAT/Model/Route.swift
```diff
@@ -99,14 +102,14 @@ class Route: NSObject, Codable {
...
+        case routeId
     }
     
     required init(from decoder: Decoder) throws {
-        
-        // MARK: #182 • Get Route ID
         let container = try decoder.container(keyedBy: CodingKeys.self)
         departureTime = Date.parseDate(try container.decode(String.self, forKey: .departureTime))
         arrivalTime = Date.parseDate(try container.decode(String.self, forKey: .arrivalTime))
+        routeId = try container.decode(String.self, forKey: .routeId)
```
Omarrasheed, line 112: Are we passing in a routeId now from the Initial backend requests that give us all the suggested routes? Just asking because if not, this may crash and we'll need some error handling

ywsang (approved):

AAAstorga (approved):
Looks good!

mattbarker016:
@ywsang, just push that last commit you made for the debug server endpoint change and we can merge this in!

Issue #175, opened 2018-10-21
Bus Alerts


mattbarker016:
# Goal

This feature will allow users to view all active alerts for the TCAT bus service. We will add a button in the About Us view that shows all active alerts (third screenshot). On the Route Options page, if any routes are affected by an active alert, we will show an indication on this page (views still in progress). The endpoint is written and active, see the copied specs below.

# Tasks

## Integrate with **/alerts** endpoint.

Some of the specs are vague, so please update the README on [ithaca-transit-backend](https://github.com/cuappdev/ithaca-transit-backend) if possible.

_Note:_ They may not currently be any active messages. If so, do your best to imitate the endpoint form and create dummy data.

- [x] Completed

## Create BusAlertsViewController to show alerts

This will be accessible from the InformationViewController. It should push in the navigation controller.

- [x] Completed

## Add logic to scan through available routes to see if any are affected by active alerts.

There may be the potential issue of an alert saying a certain route is out of service. Our backend will automatically suggest different routes, and they won't appear in the available routes. This means we technically can't catch when an active alert is affecting the results. We could potentially scan through the alert message to see if any route numbers appear, but this isn't ideal. We just may not be able to catch this scenario, but worth thinking about.

- [ ] Completed

## Create view to indicate active bus warnings.

- [ ] Completed

# Current Designs

_Initial Concepts. To be finalized on Zeplin soon._

<img width="182" alt="image" src="https://user-images.githubusercontent.com/8039928/47269190-6da8fe80-d528-11e8-8799-a9a2e1c95208.png"><img width="180" alt="image" src="https://user-images.githubusercontent.com/8039928/47269199-7a2d5700-d528-11e8-9e3b-9c0c958fc1fb.png"><img width="183" alt="image" src="https://user-images.githubusercontent.com/8039928/47269206-86b1af80-d528-11e8-9fa3-58a6632b751e.png">

## Endpoint

_Copied from backend's README. Update any helpful changes you can make on their repo!_

### **/alerts** • (GET)

**Description:** Return a list of official TCAT alerts/messages. A list should be included in the app, and periodically updated every so often. All fields are optional. 
The date, time, and daysOfWeek fields can specify an alert that takes place for example from 10 pm March 4 (fromDate) to 12 pm April 10 (toDate) on Weekdays (daysOfWeek) from 1 pm (fromTime) to 11 pm (toTime).

## Returns: [Alert]

*class* Alert

[...]

PR #194, opened 2019-02-13
Omar/service alerts

Omarrasheed:
#175


mattbarker016:
@Omarrasheed You need to resolve merge conflicts

TCAT/AppDelegate.swift
```diff
@@ -92,7 +92,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     func applicationWillEnterForeground(_ application: UIApplication) {
-        // Called as part of the transition from the background to the inactive state; here y/Users/mattbarker016ou can undo many of the changes made on entering the background.
```
mattbarker016, line 111: whoopsie

TCAT/AppDelegate.swift
```diff
@@ -115,6 +115,36 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                    if !alertRoutes.containsSameElements(as: currentRoutes) {
+                        self.userDefaults.set(alertRoutes, forKey: Constants.UserDefaults.servicedRoutes)
+                    }
+                } else {
+                    let alertRoutes = self.allRoutesInAlerts(alertRequest.data)
+                    self.userDefaults.set(alertRoutes, forKey: Constants.UserDefaults.servicedRoutes)
+                }
+            } else { print("storeServicedRoutesError: Network request failed" ) }
+        }) { (error) in
+            print("storeServicedRoutesError: \(error.localizedDescription)")
```
mattbarker016: Should we fail with a UIAlert, similar to the Couldn't Get Stops alert message? Maybe check with design

TCAT/AppDelegate.swift
```diff
@@ -115,6 +115,36 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
     }
     
+    @objc func storeServicedRoutes() {
+        Network.getAlerts().perform(withSuccess: { (alertRequest) in
+            if alertRequest.success {
+                if let currentRoutes = self.userDefaults.array(forKey: Constants.UserDefaults.servicedRoutes) as? [Int] {
+                    let alertRoutes = self.allRoutesInAlerts(alertRequest.data)
```
mattbarker016: It looks like this can be simplified to always write alertRoutes if they exist. No need to check if they're the same, you can just overwrite if they are in that case. 

```
if alertRequest.success {
    let alertRoutes = self.allRoutesInAlerts(alertRequest.data)
    self.userDefaults.set(alertRoutes, forKey: Constants.UserDefaults.servicedRoutes)
}
```

Let me know if I'm missing something here.

TCAT/AppDelegate.swift
```diff
@@ -115,6 +115,36 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                    self.userDefaults.set(alertRoutes, forKey: Constants.UserDefaults.servicedRoutes)
+                }
+            } else { print("storeServicedRoutesError: Network request failed" ) }
+        }) { (error) in
+            print("storeServicedRoutesError: \(error.localizedDescription)")
+        }
+    }
+
+    func allRoutesInAlerts(_ alerts: [Alert]) -> [Int] {
+        var allRoutes = [Int]()
```
mattbarker016: Nitpick: Could consider using `flatMap` here, no need to create a variable.

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+    func setData() {
+        if let fromDate = alert?.fromDate, let toDate = alert?.toDate {
+            timeSpanLabel.text = formatTimeString(fromDate, toDate: toDate)
+        }
+        descriptionLabel.text = alert?.message
+        if let routes = alert?.routes, !routes.isEmpty {
+            setupAffectedRoutesStackView()
+            setupaffectedRoutesLabel()
+        }
+
```
mattbarker016, line 52: Nitpick: Choose consistent spacing between blocks of code

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+        contentView.addSubview(timeSpanLabel)
+    }
+    
+    private func setupDescriptionLabel() {
+        
+        descriptionLabel = UILabel()
+        descriptionLabel.numberOfLines = 0
+        descriptionLabel.font = .getFont(.regular, size: 14)
+        descriptionLabel.textColor = Colors.primaryText
+
```
mattbarker016, line 74: Nitpick: Same as above

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+        
+        contentView.addSubview(descriptionLabel)
+    }
+    
+    private func setupaffectedRoutesLabel() {
+        
+        affectedRoutesLabel = UILabel()
+        affectedRoutesLabel.font = .getFont(.semibold, size: 18)
+        affectedRoutesLabel.textColor = Colors.primaryText
+        affectedRoutesLabel.text = "Affected Routes"
```
mattbarker016: We've started using `Constants.swift` for all client-facing strings. Add this there!

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+    }
+    
+    private func setupAffectedRoutesStackView() {
+        
+        if var routes = alert?.routes, !routes.isEmpty {
+            affectedRoutesStackView = UIStackView()
+            for _ in 0..<rowCount() {
+                var subviews = [BusIcon]()
+                for _ in 0..<maxIconsPerRow() {
+                    if !routes.isEmpty {
```
mattbarker016, line 95: You already check this on Line 81

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+                        subviews.append(BusIcon(type: .directionSmall, number: routes.removeFirst()))
```
mattbarker016: Nitpick: It's a bit cleaner to add something like `let route = routes.removeFirst()` to clearly show the step of popping the first value off; it's easy to miss buried in that line.

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+        topSeparator?.backgroundColor = Colors.backgroundWash
+        
+        contentView.addSubview(topSeparator!)
+    }
+    
+    override func updateConstraints() {
+        
+        if let topSeparator = topSeparator {
+            topSeparator.snp.remakeConstraints { (make) in
+                make.top.leading.trailing.equalToSuperview().labeled("topSeparator: Top, Leading, Trailing")
```
mattbarker016: I've never seen this syntax! What does it do / what are the benefits over the usual way we use SnapKit?

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+            descriptionLabel.snp.makeConstraints { (make) in
+                make.bottom.equalToSuperview().inset(borderInset).labeled("descriptionLabel: Bottom")
+            }
+        }
+        super.updateConstraints()
+    }
+    
+    private func formatTimeString(_ fromDate: String, toDate: String) -> String {
+        
+        // ISSUE IN PARSING STRING (HITS DEFAULT DATE.DISTANTPAST)
```
mattbarker016: Resolve whatever this issue is referring to. Also, check out some of the Time.swift functions to see if we possibly already do something like this. Edit: I see the modification you made in an Extension, still would like to hear the issue

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+        formatter.dateFormat = "EEEE M/d"
+
+        if let unWrappedFromDate = fromDate, let unWrappedToDate = toDate {
+            let formattedFromDate = formatter.string(from: unWrappedFromDate)
+            let formattedToDate = formatter.string(from: unWrappedToDate)
+            
+            return "\(formattedFromDate) - \(formattedToDate)"
+        }
+        
+        return "Time: Unknown"
```
mattbarker016, line 222: Maybe check in with design on a better default message, perhaps even an empty string

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+    }
+    
+    private func maxIconsPerRow() -> Int {
+        let iconWidth = Int(BusIconType.directionSmall.width)
+        let screenWidth = Int(UIScreen.main.bounds.width)
+        let minSpacing = 10
```
mattbarker016: Is this used elsewhere in this file? Should be top level in case changed

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+        let totalConstraintInset = borderInset * 2
+        
+        return (screenWidth - totalConstraintInset + minSpacing) / (iconWidth + minSpacing)
+    }
+    
+    private func rowCount() -> Int {
+        guard let routes = alert?.routes else { return 0 }
+        
+        if routes.count > maxIconsPerRow() {
```
mattbarker016: Consider storing the result of `maxIconsPerRow()` into a variable so you don't have to keep recalculating it

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,238 @@
...
+        if let stackView = affectedRoutesStackView {
+            stackView.removeFromSuperview()
+        }
+        affectedRoutesStackView = nil
+        
+        if let routesLabel = affectedRoutesLabel {
+            routesLabel.removeFromSuperview()
+        }
+        affectedRoutesLabel = nil
+    }
```
mattbarker016, line 255: Nitpick: Add space

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,175 @@
+//
+//  ServiceAlertsViewController.swift
+//  TCAT
+//
+//  Created by Omar Rasheed on 10/23/18.
+//  Copyright © 2018 cuappdev. All rights reserved.
+//
+import UIKit
+import SnapKit
+import DZNEmptyDataSet
```
mattbarker016, line 10: Nitpick: Add space

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+                tableView.scrollToRow(at: .init(row: 0, section: 0), at: .top, animated: false)
+            }
+        }
+    }
+    var priorities = [Int]()
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        title = "TCAT Service Alerts"
```
mattbarker016: Move to `Constants.swift`

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+        
+        view.backgroundColor = Colors.backgroundWash
+        tableView = UITableView(frame: .zero, style: .grouped)
+        tableView.backgroundColor = view.backgroundColor
+        tableView.dataSource = self
+        tableView.delegate = self
+        tableView.register(ServiceAlertTableViewCell.self, forCellReuseIdentifier: ServiceAlertTableViewCell.identifier)
+        tableView.allowsSelection = false
+        
+        // Any idea why I'm needing to do this??
```
mattbarker016: Why do you? ¯\_(ツ)_/¯

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+    
+    func getServiceAlerts() {
+        Network.getAlerts().perform(withSuccess: { (request) in
+            if (request.success) {
+                self.alerts = self.sortedAlerts(alertsList: request.data)
+            }
+        }) { (error) in
+            let fileName = "ServiceAlertsVieController"
```
mattbarker016: Again, consider showing a UIAlert error

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+        return sortedAlerts
+    }
+}
+
+extension ServiceAlertsViewController: UITableViewDelegate {
+    
+    func tableView(_ tableView: UITableView, viewForHeaderInSection section: Int) -> UIView? {
+        let headerView = HeaderView()
+        
+        switch priorities[section] {
```
mattbarker016, line 174: Did you swap these from low to high like we discovered?

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+    func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        return UITableView.automaticDimension
+    }
+    
+    func numberOfSections(in tableView: UITableView) -> Int {
+        return priorities.count
+    }
+    
+    func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        if let alerts = alerts[priorities[section]] {
```
mattbarker016: `return alerts?[priorities[section]].count ?? 0`

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -0,0 +1,175 @@
...
+        let cell = tableView.dequeueReusableCell(withIdentifier: ServiceAlertTableViewCell.identifier, for: indexPath) as? ServiceAlertTableViewCell
+        
+        if let alertList = alerts[priorities[indexPath.section]] {
+            cell?.alert = alertList[indexPath.row]
+            cell?.rowNum = indexPath.row
+            cell?.setData()
+            cell?.setNeedsUpdateConstraints()
+        }
+        
+        return cell!
```
mattbarker016: Hmm, not a fan of `!`. Is there a version of `dequeueReusable` that isn't optional?

mattbarker016 (changes requested):
Meaty PR! I've yet to build this on device and really mess around, but let's get a round of fixes in first. As always, even though I'm often just saying things to do and nitpicks, good code overall and you're that much closer to a feature shipping!

TCAT/AppDelegate.swift
Omarrasheed: Oh this stuff was actually for keeping information about alerts globally to later present but I don't think we need to deal with that on this pr. We kinda scratched a lot of the ideas about presenting them around the app. Just took it out thanks for catching that.

TCAT/Cells/ServiceAlertTableViewCell.swift
Omarrasheed: Helps a lot with debugging. Whenever there are constraint errors or warnings, instead of giving a constraint identifier which is a weird number attached to a view, it prints the label you pass in instead. I'll take them out tho, they def make the code look more cluttered.

TCAT/Cells/ServiceAlertTableViewCell.swift
Omarrasheed: I used this line cuz i keep popping from the list when I add each subview

TCAT/Cells/ServiceAlertTableViewCell.swift
```diff
@@ -0,0 +1,237 @@
...
+            let formattedFromDate = formatter.string(from: unWrappedFromDate)
+            let formattedToDate = formatter.string(from: unWrappedToDate)
+            
+            return "\(formattedFromDate) - \(formattedToDate)"
+        }
+        
+        return "Time: Unknown"
+    }
+    
+    private func calculateMaxIconsPerRow() -> Int {
```
mattbarker016: Maybe this should just be a computed property, but then you have to recalculate it each time. @AAAstorga Thoughts on making a layout-based calculation like this be a function or computed property? If you need more context, poke around or ask

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -121,24 +121,22 @@ extension ServiceAlertsViewController: UITableViewDataSource {
...
-        } else {
-            return 0
-        }
+        return alerts[priorities[section]]?.count ?? 0
     }
     
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = tableView.dequeueReusableCell(withIdentifier: ServiceAlertTableViewCell.identifier, for: indexPath) as? ServiceAlertTableViewCell
+        
+        var cell : ServiceAlertTableViewCell!
```
mattbarker016: @drewsdunne @AAAstorga Do we have a good standard way to do this? I mentioned before, I never like using `!`

mattbarker016:
I added one or two extra comments, are they appearing?

TCAT/Controllers/ServiceAlertsViewController.swift
Omarrasheed: Could guard let? And if it fails, we return the base UITableViewCell (so the program doesn't crash for the user) in addition to showing some sort of error message (either alert or just print some error message so we can debug on our side)

mattbarker016: I like that!

TCAT/Controllers/ServiceAlertsViewController.swift
```diff
@@ -126,8 +202,7 @@ extension ServiceAlertsViewController: UITableViewDataSource {
     
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         
-        var cell : ServiceAlertTableViewCell!
-        cell = tableView.dequeueReusableCell(withIdentifier: ServiceAlertTableViewCell.identifier) as? ServiceAlertTableViewCell
+        guard let cell = tableView.dequeueReusableCell(withIdentifier: ServiceAlertTableViewCell.identifier) as? ServiceAlertTableViewCell else { return UITableViewCell() }
```
mattbarker016: nitpick, put else on second line

TCAT/Supporting Files/Constants.swift
```diff
@@ -14,11 +14,18 @@ struct Constants {
...
         struct GeneralActions {
             static let settings = "Settings"
             static let cancel = "Cancel"
             static let dontRemind = "Don't Remind Me Again"
         }
+        
+        struct AlertsRequestFailure {
+            static let title = "Request Failed"
+            static let message = "We failed to retrieve alerts. Pull to retry"
+            static let action = "OK!"
```
mattbarker016: no "!"

TCAT/Supporting Files/Constants.swift
```diff
@@ -14,11 +14,18 @@ struct Constants {
...
 
+        
         struct GeneralActions {
             static let settings = "Settings"
             static let cancel = "Cancel"
             static let dontRemind = "Don't Remind Me Again"
         }
+        
+        struct AlertsRequestFailure {
+            static let title = "Request Failed"
```
mattbarker016: title: "Couldn't Fetch Service Alerts"

TCAT/Supporting Files/Constants.swift
```diff
@@ -14,11 +14,18 @@ struct Constants {
...
+            static let message = "We failed to retrieve alerts. Pull to retry"
```
mattbarker016: "There was an error fetching service alerts. Please retry again."

TCAT/Supporting Files/Constants.swift
```diff
@@ -136,6 +143,7 @@ struct Constants {
         static let locationNotFound = "Location Not Found"
         static let noNetworkConnection = "No Network Connection"
         static let noRoutesFound = "No Routes Found"
+        static let noActiveAlerts = "No Active Alerts. Check Back Later!"
```
mattbarker016: "No Active Service Alerts"

mattbarker016 (approved):
LGTM. Would like to just play with on TestFlight, then good to ship

Omarrasheed:
When testing service alerts:
        1) Comment out getServiceAlerts at the bottom of viewDidLoad()
        2) Add the line `createDummyData()` below it

If there are any cases you would like to test, initialize another alert object on line 282 and append it to alertData before the last line

Issue #196, opened 2019-02-18
Refactor PlaceResult & BusStop into one full Place class


mattbarker016:
PlaceResult and BusStop are subclasses of `Place`, a very basic protocol-esque class that both location-based data structures share. Currently, PlaceResult does not store actual coordinates, but rather a `placeId` that hashes to a certain location via Google lookup magic. This extra async lookup is costly, annoying, and redundant.

We will create a new `Place` class that stores the metadata and coordinates of all locations. The first time a Google Place is selected or utilized, we will look up the coordinates via `placeId` **once** and store them on device for local usage, favorites, etc.

Hopefully this doesn't tear up too much existing stuff, but this process needs to be made simpler.

Issue #197, opened 2019-02-18
Implement /search endpoint


mattbarker016:
Documentation can be found [here](https://github.com/cuappdev/ithaca-transit-backend#search). This feature was added based on this [issue](https://github.com/cuappdev/ithaca-transit-backend/issues/101).

This issue includes removing client-side logic dedicated to sorting by bus and Google Place, and potentially the Fuzzywuzzy_swift autocomplete dependency. We should try and simplify as much as possible.

PR #198, opened 2019-02-27
Place Model Refactoring (#196, #197)

mattbarker016:
This PR is still a work in progress (see **Improvements** section below), but would appreciate starting to get some eyeballs on this. There's a lot to read, but it should provide a lot of context for what's going on here, so please at least skim!

# Background

Previously, we used `BusStop` and `PlaceResult`, each of which were subclasses of a close to empty `Place` model, to handle the various locations used within the app. This led to a lot of overhead involving switch statements, multiple UITableViewCells, etc. all basically doing the same thing. This implementation was used because Google Places requires a separate network call to fetch coordinates based on an identifier.

# Place.swift

The new `Place` model has the following structure.
- `name: String` • the name of the place
- `type: PlaceType` • the type of the place (`.busStop` | `.googlePlace` | `.unknown`)
- `description: String` • a description of the place ("South Meadow Street, Ithaca NY", "Bus Stop")
- `placeIdentifier: String?` • a unique identifier for the place (used for `.googlePlace`)
- `latitude` : Double?
- `longitude` : Double?

## Data Migration

There is a function `migrationToNewPlacesModel()` in `AppDelegate.swift` that converts the stored recent searches and favorites to the new model. The function accesses the existing data using `NSKeyedUnarchiver.unarchiveObject(with: dataFromUserDefaults) as? [Any] ` and runs `convertDataToPlaces`. This takes in `[Any]`, casts to `BusStop` and `PlaceResult`, and then moves properties to the `Place` model, as well as looks up coordinates via `CoordinateVisitor` if needed. The array of places is then encoded with `JSONEncoder` and stored in the same key. Based on `if let` statements and type casting, these functions will halt once the new data model is saved. One possible improvement would be to limit the migration function to only fire once based on the version number of the app.

## Using Place.swift

The coordinate values (`latitude`, `longitude`) are optional because, in some cases, the information is not immediately available and needs to be requested. This is done using `CoordinateVisitor` whenever a Google Place is requested, which occurs in the `FavoritesTableViewController` and `SearchResultsTableViewController`. This should be written as few times as possible and can still be improved.

# /search Endpoint

Previously, on the client we merged Google Places and bus stops into one search results list with sorting. This logic has been moved to the backend, which now conforms to a model where all search results have a `name` and `type` and either bus stop coordinates or Google Place information. `Place`  conforms to `Codable` and `SearchResult` processes this array of results.

# PlaceTableViewCell.swift

With a simplified `Place` model, the UITableViewCell classes `BusStopCell` and `SearchResultsCell` were able to be merged into one definitive cell for places. This class has a variable called `place: Place?` that should be set to the relevant place the cell is representing after calling `tableView.dequeueReusableCell(...)`. This will change the visual layout of the cell (e.g. pin color) appropriately. This process could potentially be improved better.

# Other Odds & Ends
Here is a list of other changes made and things touched.

- Removed all functionality and references to "Cornell Destinations", a v0 idea.
- Banner fixes started regarding excessive haptic feedback and looping

## Improvements
Below are known issues or improvements that can be made.

### Important

- [X] Converted data doesn't show on first load, need to go to another screen and back.
- [X] Improving data migration process to only strictly fire once (see **Data Migration** section)
- [X] Improved error handling for data conversion, encoding, decoding aspects. Set up Fabric to report any failures to diagnose any migration issues.
- [X] Fixing `imageView` in `PlaceTableViewCell` not appearing until scrolled past in `AllStopsTableViewController`
- [X] Current Location Cell not properly displaying in SearchResultsTableViewController
- [X] VersionStore TODO: Removing a side-effect of a getter function
- [X] Favorites TVC TODO: Proper error handling for failing CoordinateVisitor

### To Improve in Future PR

- [ ] Standardizing `JSONEncoder` / `JSONDecoder` usage throughout app. This can be done by having one reference to an `encoder` and `decoder`. `userDefaults` should follow a similar pattern.
- [ ] There are some general instances where route-related logic can be further simplified and moved to backend. Look for TODO comments.
- [ ] Haptic feedback banner issue!


mattbarker016:
This will fully resolve #196 and #197

TCAT/AppDelegate.swift
```diff
@@ -30,6 +31,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         
+        // Set Up Google Services
+        FirebaseApp.configure()
+        GMSServices.provideAPIKey(Keys.googleMaps.value)
+        GMSPlacesClient.provideAPIKey(Keys.googlePlaces.value)
+        
+        migrationToNewPlacesModel()
```
doctor3w: I think adding a check here for some migration key would be ideal, rather than in the function itself. I would be cautious about using the version number because someone could feasibly jump multiple versions. Storing version number of the last migration may suffice because then one could check if the last migration was up to date, in case we ever need to migrate again. For example, v1.0.1 we add first migration from v1.0.0, then in v1.0.3 we need an even newer migration, check if version is at least something but less than latest migration. I'm sure you can think of specifics.

TCAT/Cells/PlaceTableViewCell.swift
```diff
@@ -8,12 +8,15 @@
 
 import UIKit
 
-class SearchResultsCell: UITableViewCell {
-    let labelWidth: CGFloat = 45.0
+class PlaceTableViewCell: UITableViewCell {
+    
+    var place: Place?
```
doctor3w: I'm hesitant about giving the cell view a reference to the model it shows, is storing it here necessary?

TCAT/Controllers/HomeViewController.swift
```diff
@@ -320,35 +315,21 @@ extension HomeViewController: UITableViewDataSource {
...
-                cell.imageView?.image = #imageLiteral(resourceName: "list")
-                cell.accessoryType = .disclosureIndicator
-            }
+        if sections[indexPath.section].type == .seeAllStops {
+            cell = tableView.dequeueReusableCell(withIdentifier: Constants.Cells.seeAllStopsIdentifier)
+            cell.textLabel?.text = Constants.General.seeAllStops
+            cell.imageView?.image = #imageLiteral(resourceName: "list")
+            cell.accessoryType = .disclosureIndicator
+        }
+
```
doctor3w: I've never seen this kind of formatting for an if-else stmt. It's...unique.

TCAT/Utilities/Endpoints.swift
```diff
@@ -128,24 +113,22 @@ class Network {
...
-        let request: APIRequest<JSON, Error> = tron.swiftyJSON.request("places")
+    
+    class func getSearchResults(searchText: String) -> APIRequest<SearchRequest, Error> {
+        let request: APIRequest<SearchRequest, Error> = tron.codable.request("search")
         request.method = .post
-            request.parameterEncoding = JSONEncoding.default
+        request.parameterEncoding = JSONEncoding.default
         request.parameters = [
-            "query": searchText
+            "query" : searchText
```
doctor3w: Remove whitespace before ':'

TCAT/Utilities/SearchTableViewHelpers.swift
```diff
@@ -51,172 +47,79 @@ class SearchTableViewManager {
...
-                    if let busStop = place as? BusStop {
-                        itemTypes.append(.busStop(busStop))
-                    }
-                    if let searchResult = place as? PlaceResult {
-                        itemTypes.append(.placeResult(searchResult))
-                    }
-
-                }
-                return itemTypes
+            if let places =  try? decoder.decode([Place].self, from: storedPlaces) {
```
doctor3w: can't you chain these if's using a comma?

TCAT/Utilities/SearchTableViewHelpers.swift
```diff
@@ -51,172 +47,79 @@ class SearchTableViewManager {
...
-                return busStop.name
-            }
-            if let place = location as? PlaceResult {
-                return place.name
-            }
-            return ""
+    
+        places = bottom ? places + [place] : [place] + places
+        
+        if places.count > limit { places.remove(at: places.count - 1)}
```
doctor3w: Add a whitespace before final }

doctor3w (commented):
Love the work! This looks like a solid change. Pretty huge PR, I'm not sure if this chunked up before but it would be nice if along the way you PR'd some work, got reviews, made changes. Great work though, looks good all around. Just had a few changes and questions.

TCAT/AppDelegate.swift
```diff
@@ -105,6 +108,104 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        
+        //
+        // favorites
+        //
+        
+        let favoritesKey = Constants.UserDefaults.favorites
+        
+        if let storedPlaces = userDefaults.value(forKey: favoritesKey) as? Data {
+            // This will only fire on legacy models
+            if let favorites = NSKeyedUnarchiver.unarchiveObject(with: storedPlaces) as? [Any] {
```
Omarrasheed: nitpick: can we make this one if let like `if let storedPlaces = ___, let favorites = ____ { ... }`

TCAT/AppDelegate.swift
```diff
@@ -105,6 +108,104 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        
+        //
+        // recentSearches
+        //
+        
+        let recentSearchesKey = Constants.UserDefaults.recentSearch
+        
+        if let storedPlaces = userDefaults.value(forKey: recentSearchesKey) as? Data {
+            //  This will only fire on legacy models
+            if let recents = NSKeyedUnarchiver.unarchiveObject(with: storedPlaces) as? [Any] {
```
Omarrasheed: Same as above

TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -161,26 +161,28 @@ class AllStopsTableViewController: UITableViewController {
     }
 
     override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = tableView.dequeueReusableCell(withIdentifier: Constants.Cells.busIdentifier, for: indexPath) as! BusStopCell
+        let cell = tableView.dequeueReusableCell(withIdentifier: Constants.Cells.placeIdentifier, for: indexPath) as! PlaceTableViewCell
```
Omarrasheed: More of a general comment but should we go into all of the cellForRowAt functions that have this force casting for cells and change them to the `guard let cell = dequeue... else { return UITableViewCell() }`

Omarrasheed (commented):
Only have like two or three small nitpick comments but LGTM. Really phat pr lol I'm glad to see that this really simplified a lot of our stuff in that tableview and just overall organization of our different data types. Def helps w codable parsing too

TCAT/Controllers/AllStopsTableViewController.swift
doctor3w: This actually is not necessary as this should always cast properly.

TCAT/Cells/PlaceTableViewCell.swift
mattbarker016: @drewsdunne At the moment I only use it so auto configure what color for the pin based on whether it's a bus stop or not. The cell is the same except slight aesthetic differences based on the type of the place, what do you recommend?

doctor3w: I usually just provide a function such as setup(for place: Place) and call that tableView function but neither is really a great MVC solution :( so whatever works.

PR #199, opened 2019-03-01
Omar/whats new fix

Omarrasheed:


TCAT/Views/WhatsNewHeaderView.swift
```diff
@@ -78,65 +84,84 @@ class WhatsNewHeaderView: UIView {
...
-            make.top.equalTo(updateTitle.snp.bottom).offset(6)
-            make.centerX.equalToSuperview()
+            updateDescToUpdateName = make.top.equalTo(updateTitle.snp.bottom).offset(6).constraint
             make.leading.trailing.equalToSuperview().inset(value)
-            if
-                let containerView = superview,
-                let tableView = containerView.superview,
-                let wholeView = tableView.superview,
-                let description = updateDescription.text
+            if let description = updateDescription.text
```
mattbarker016: nit pick, put `{` on `if let` line

TCAT/Views/WhatsNewHeaderView.swift
```diff
@@ -78,65 +84,84 @@ class WhatsNewHeaderView: UIView {
...
-            make.height.equalTo(dismissButton.intrinsicContentSize.height)
+            dismissButtonToBottom = make.bottom.equalToSuperview().inset(16).constraint
         }
-
-        snp.makeConstraints { (make) in
-            make.bottom.equalTo(dismissButton.snp.bottom).offset(16)
+    }
+    
+    func calculateCardHeight() -> CGFloat {
+        if let titleToTop = titleToTop, let updateNameToTitle = updateNameToTitle, let updateDescToUpdateName = updateDescToUpdateName, let dismissButtonToUpdateDesc = dismissButtonToUpdateDesc, let dismissButtonToBottom = dismissButtonToBottom {
```
mattbarker016: nitpick: break up `let` statements on multiple lines

```
if
    let x = ...,
    let y = ...
{
    ...
}
```

TCAT/Views/WhatsNewHeaderView.swift
```diff
@@ -78,65 +84,84 @@ class WhatsNewHeaderView: UIView {
...
+            
+            let titleToTopVal = titleToTop.layoutConstraints[0].constant
```
mattbarker016: is this a safe value to access? Can this ever fail?

TCAT/Views/WhatsNewHeaderView.swift
```diff
@@ -78,65 +84,84 @@ class WhatsNewHeaderView: UIView {
...
+            let updateDescToUpdateNameVal = updateDescToUpdateName.layoutConstraints[0].constant
+            
+            let updateDescSpace = updateDescToUpdateNameVal + updateDescriptionHeight
+            
+            let dismissButtonToUpdateDescVal = dismissButtonToUpdateDesc.layoutConstraints[0].constant
+            let dismissButtonHeight = dismissButton.intrinsicContentSize.height
+            
+            let dismissButtonSpace = dismissButtonToUpdateDescVal + dismissButtonHeight
+            
+            let bottomOffset = -dismissButtonToBottom.layoutConstraints[0].constant
```
mattbarker016: appreciate the named variables, readable

mattbarker016 (approved):
LGTM, again just wanna see on a build via TestFlight

TCAT/Views/WhatsNewHeaderView.swift
Omarrasheed: With how I've structured it, It shouldn't because if its not nil, then there exists a constraint and layoutConstraints[0] will always have something to access

mattbarker016:
In this ongoing PR can you standardize the process of changing What's New information better? You'll work with the VersionStore boolean logic to have the card show anytime we change the respective Constants.swift file instead of relying on `1.x` minor version changes. I think we'll just need to store some identifier of the message in UserDefaults.

A possible idea could be creating a second button in the card for action items (primary, secondary) instead of a hyperlink in the message, but we'll run that by design.

PR #202, opened 2019-03-06
What's New Card Improvements

mattbarker016:
image

I improved the **What's New** internal model, to make it simple to add and update the What's New feature's data and links. Now, you can create **up to two buttons** (primary and secondary), and based on whether there is data or not, the view will update accordingly. I tested edge cases for the `.current` variable (two buttons, one primary, one secondary), and made sure the web link works, but not the Twitter link.


Omarrasheed (approved):
Looks great!

PR #208, opened 2019-03-14
Change dev server 

ywsang:
Point to https://transit-testflight.cornellappdev.com


kevinchan159 (approved):
beautiful

mattbarker016:
We only need to worry about version and build number on release, but good to keep in mind!

mattbarker016 (approved):

PR #213, opened 2019-03-30
Add out of range location error message

ywsang:
Added a check to the Today Extension to display "Location Out of Range" when the user is not within Tompkins Country. Previously the loading indicator would spin indefinitely.


Omarrasheed (approved):
Nice catch

kevinchan159 (approved):
a m a z i n g

mattbarker016:
Nice! On mobile, but just add a space for one of the if statements you did after the parentheses. Good to merge otherwise!

Issue #215, opened 2019-04-10
SWIFT 5 UPDATES


Omarrasheed:
First step:
- [x] Update project to swift 5

Second step:
- [x] Update pods
- [x] Swiftlint

PR #216, opened 2019-04-14
Omar/swift5

Omarrasheed:
#215 
- Migrate to Swift 5
- Remove unnecessary pods
- Run Swiftlint Autocorrect, add it to build script, and fix some of the easier warnings
- Will tackle the rest of the warnings in another issue next semester

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -18,16 +18,17 @@ excluded: # paths to ignore during linting. Takes precedence over `included`.
 force_cast: warning # implicitly. Give warning only for force casting
 
 force_try:
-severity: warning # explicitly. Give warning only for force try
+  severity: warning # explicitly. Give warning only for force try
 
 type_body_length:
 - 300 # warning
-- 400 # error
 
 # or they can set both explicitly
 file_length:
-warning: 500
-error: 800
+- 500
+
+line_length:
+- 120
 
 large_tuple: # warn user when using 3 values in tuple, give error if there are 4
 - 3
--- a/Podfile
+++ b/Podfile
@@ -33,6 +33,9 @@ target 'TCAT' do
     pod 'Presentation', :git=> 'https://github.com/cuappdev/Presentation.git'
     pod 'SnapKit'
     pod 'WhatsNewKit', '~> 1.1.2'
+
+    # Other
+    pod 'SwiftLint'
     
     target 'TCATTests' do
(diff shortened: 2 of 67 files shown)


ywsang (approved):

.swiftlint.yml
```diff
@@ -18,16 +18,17 @@ excluded: # paths to ignore during linting. Takes precedence over `included`.
 force_cast: warning # implicitly. Give warning only for force casting
 
 force_try:
-severity: warning # explicitly. Give warning only for force try
+  severity: warning # explicitly. Give warning only for force try
```
mattbarker016, line 21: Should this be on the same line, or is it like an if block?

mattbarker016:
Were there any big changes other than spacing stuff? A bit hard to parse 😅

Omarrasheed:
Not rly lmao most of it was swiftlint autocorrect. The main things r that I changed the proj to be on swift5 now. Migration was smooth and no errors. There were like a depracated function in extensions for nsrange I think but that was about the extent of those fixes. I also took out fuzzy_wuzzy swift

.swiftlint.yml
Omarrasheed: I think this can be on the same line. I think this was taken from their demo code and they were showing the different ways the settings can be changed. I'll put it on the same line

mattbarker016: Was just curious!

mattbarker016:
Did you make these changes yourself, or it did them all?

```
    func nsRange(from range: Range<String.Index>) -> NSRange {
	        let from = range.lowerBound.encodedOffset
	        let to = range.upperBound.encodedOffset
	        return NSRange(location: from - startIndex.encodedOffset, length: to - from)
	        let from = range.upperBound.utf16Offset(in: self)
	        let to = range.upperBound.utf16Offset(in: self)
	        return NSRange(location: from - startIndex.utf16Offset(in: self), length: to - from)
	    }
```

Omarrasheed:
I had to change them manually

mattbarker016:
One should be lowerBound and one should be upperBound. Look at the original code.

Omarrasheed:
Good catch thank you!

mattbarker016:
If you think there could be any other sneaky little changes like that, give it one more once over. Otherwise, looks good to me.

Issue #212, opened 2019-03-27
Show buses delayed by < 120 seconds as On Time


ywsang:
In order to accurately display delayed buses, we need to overcome the fact that live tracking data is only updated every 30 seconds. 

In order to do this: 
- Show buses that are delayed by < 120 seconds as "On Time", EXCEPT if the actual departTime + delay > current time.

PR #218, opened 2019-04-17
Change delay logic for buses delayed for < 120 seconds

ywsang:
Addresses #212.

### DIFF
--- a/TCAT/Cells/RouteTableViewCell.swift
+++ b/TCAT/Cells/RouteTableViewCell.swift
@@ -264,6 +264,7 @@ class RouteTableViewCell: UITableViewCell {
                 
                 let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
                 
+                // Our live tracking only updates once every 30 seconds, so we want to show buses that are delayed by < 120 as on time in order to be more accurate about the status of slightly delayed buses. This way riders get to a bus stop earlier rather than later when trying to catch such buses.
                 if Time.compare(date1: departTime, date2: delayedDepartTime) == .orderedAscending { // bus is delayed
                     if (delayedDepartTime >= Date() || delay >= 120) {
                         return .late(date: delayedDepartTime)
--- a/TCAT/Cells/RouteTableViewCell.swift
+++ b/TCAT/Cells/RouteTableViewCell.swift
@@ -264,10 +264,14 @@ class RouteTableViewCell: UITableViewCell {
                 
                 let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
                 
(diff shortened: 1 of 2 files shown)


Omarrasheed (approved):
lgtm

TCAT/Cells/RouteTableViewCell.swift
```diff
@@ -264,10 +264,14 @@ class RouteTableViewCell: UITableViewCell {
                 
                 let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
                 
-                if Time.compare(date1: departTime, date2: delayedDepartTime) == .orderedAscending {
-                    return .late(date: delayedDepartTime)
+                if Time.compare(date1: departTime, date2: delayedDepartTime) == .orderedAscending { // bus is delayed
+                    if (delayedDepartTime >= Date() || delay >= 120) {
+                        return .late(date: delayedDepartTime)
+                    } else { // delay < 120
+                        return .onTime(date: departTime) // what time??
```
kevinchan159: what does the `// what time?` comment mean

kevinchan159 (approved):

TCAT/Cells/RouteTableViewCell.swift
ywsang: whoops it was for when I was asking you what time to display in that case, I just removed it

mattbarker016 (commented):
Just to confirm, we're technically going to show buses as time even when they're up 2 min late?

I'm sorry to beat a dead horse a bit, but can you make an extra explicit comment in here that explains what this means? "If buses are a bit late, they possibly won't show in the route list as a valid bus to take, even though it might catch up." This check makes sure buses are definitively late.

Is there a reason we considered 120 over 60 sec also?

kevinchan159:
> Just to confirm, we're technically going to show buses as time even when they're up 2 min late?
> 
> I'm sorry to beat a dead horse a bit, but can you make an extra explicit comment in here that explains what this means? "If buses are a bit late, they possibly won't show in the route list as a valid bus to take, even though it might catch up." This check makes sure buses are definitively late.
> 
> Is there a reason we considered 120 over 60 sec also?

@mattbarker016 The reason is because our live tracking only updates once every 30 seconds which means that within a 60 second timespan, it could be that the location only gets updated once. By having a greater threshold, we'll be more accurate about this. Plus, this means that riders will always get to the stop earlier than later which is better than missing the bus

ywsang:
@mattbarker016 An explicit comment in the code itself?

mattbarker016:
@ywsang I would include the first part of Kevin's above comment in the code. @kevinchan159 So why not making the constant 60 sec?

Issue #229, opened 2019-04-30
Siri Shortcuts Fails to Open to RouteOptions


ywsang:
On newly created Siri shortcuts, the app fails to open to RouteOptions VC. It looks like it's because the `latitude` and `longitude` properties of `Place` are now optional, so converting the coordinates to strings to create the URL scheme results in "Optional(some #)". Thus, the URL fails to parse correctly when trying to open the app via the URL scheme.

PR #232, opened 2019-05-02
Fix Siri Shortcut Donations

ywsang:
This PR addresses issue #229.
The donation process in `RouteOptions` was fixed to unwrap the coordinate values before converting to Strings. This PR also checks the app version to delete any shortcut donations that have improperly formed coordinates.

### DIFF
--- a/TCAT/AppDelegate.swift
+++ b/TCAT/AppDelegate.swift
@@ -79,7 +79,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let navigationController = showOnboarding ? OnboardingNavigationController(rootViewController: rootVC) :
             CustomNavigationController(rootViewController: rootVC)
         
-        // v1.3 Data Migration
+        patchFunctions(rootVC: rootVC)
+
+        // Initalize window without storyboard
+        self.window = UIWindow(frame: UIScreen.main.bounds)
+        self.window!.rootViewController = navigationController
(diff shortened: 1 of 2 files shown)


TCAT/AppDelegate.swift
```diff
@@ -92,6 +93,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 Analytics.shared.log(payload)
             }
         }
+        
+        // v1.4.1 Delete Corrupted Shortcut Donations
```
mattbarker016, line 108: Can you just put all the patch functions in a `patchFunctions()` wrapper and get it out of that first AppDel function?

ywsang: I tried doing that, but I wasn't sure how to pass the `rootVC` that the first path (v1.3 Data Migration) needs

mattbarker016: Oo right. You can pass it as a parameter maybe? I'll defer to your judgement of whether to keep both functions in there, abstract out, etc. Whatever you think is best!

ywsang: I'll try out passing it as a parameter later today -- if it gets complicated I think it'd be best just to leave both in there

mattbarker016:
LooksGoodToMe

Omarrasheed (approved):
LGTM

PR #233, opened 2019-05-05
Fixed bug with travel time wasn't showing properly

Omarrasheed:


kevinchan159 (approved):
speedy

mattbarker016:
What was the bug here?

Omarrasheed:
Old behavior: Was displaying the end time incorrectly in the routeOptionsView. It was doing some extra filtering on the returned routes and but it was not grabbing it correctly when there was a walking at the very end.

ywsang (approved):
lgtm

mattbarker016 (approved):

Issue #242, opened 2019-05-11
InformationVC: Make the table scrollable


Omarrasheed:
On smaller phones, we have the issue of users not being able to scroll down to see all of the fields in our tableView. Example:

<img src="https://user-images.githubusercontent.com/36868927/57575476-b9fe5200-7418-11e9-9fa8-5c188a618f0f.png" width="300">

PR #247, opened 2019-05-19
Make Information VC scrollable

Omarrasheed:
#242 

Fixed this by putting all of the views that were previously constrained above the tableview into a containerView that was used as the headerView of the tableView

### DIFF
--- a/TCAT/Controllers/InformationViewController.swift
+++ b/TCAT/Controllers/InformationViewController.swift
@@ -10,19 +10,20 @@ import MessageUI
 import SafariServices
 import UIKit
 
-class InformationViewController: UIViewController, UITableViewDataSource, UITableViewDelegate, MFMailComposeViewControllerDelegate {
+class InformationViewController: UIViewController {
 
-    private var appDevImage = UIImageView()
-    private var appDevTitle = UILabel()
-    private var descriptionLabel = UILabel()
-    private var dismissButton = UIButton(type: .system)
-    private var hiddenLabel = UILabel()
-    private var sendFeedbackButton = UIButton()
-    private var tableView = UITableView(frame: .zero, style: .grouped)
-    private var tcatImage = UIImageView()
-    private var tcatQuoteText = UILabel()
-    private var titleLabel = UILabel()
-    private var visitWebsiteButton = UIButton()
+    private let appDevImage = UIImageView()
+    private let appDevTitle = UILabel()
+    private let descriptionLabel = UILabel()
+    private let dismissButton = UIButton(type: .system)
+    private var headerView: UIView!
+    private let hiddenLabel = UILabel()
+    private let sendFeedbackButton = UIButton()
+    private let tableView = UITableView(frame: .zero, style: .grouped)
+    private let tcatImage = UIImageView()
+    private let tcatQuoteText = UILabel()
+    private let titleLabel = UILabel()
+    private let visitWebsiteButton = UIButton()
 
     private var content: [[(name: String, action: Selector)]] = [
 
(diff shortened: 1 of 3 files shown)


TCAT/Controllers/InformationViewController.swift
```diff
@@ -42,138 +43,137 @@ class InformationViewController: UIViewController, UITableViewDataSource, UITabl
...
+        tableView.showsVerticalScrollIndicator = false
+
+        setupHeaderView()
+        tableView.tableHeaderView = headerView
+
         view.addSubview(tableView)
+    }
+
+    func setupHeaderView() {
+        headerView = UIView(frame: CGRect(x: 0, y: 0, width: headerWidth, height: headerHeight))
```
kevinchan159: i think it might be better to just create a custom UIView `InformationHeaderView` and then you just use that inside here instead of having all the views in here as well. want to avoid big view controllers as much as possible

kevinchan159 (changes requested):
great work! just left 1 comment

mattbarker016 (approved):
+1 to Kevin’s comment

ywsang (commented):
looks good! agree with kevin's comment as well

TCAT/Controllers/InformationViewController.swift
```diff
@@ -42,138 +43,137 @@ class InformationViewController: UIViewController, UITableViewDataSource, UITabl
 
     ]
 
+    private let descriptionLabelSize = CGSize(width: 167, height: 34)
+    private let descriptionLabelTopOffset: CGFloat = 12
+    private let headerWidth = UIScreen.main.bounds.width
+    /** Represents the total height of the headerView given the width of the phone screen. Most of the variables
+     used are static except for the tcat image height, which is dependent on the width of the user's screen.
+     The extra 37 is used for bottom inset padding for the headerView, but I felt it would be unnecessary to
+     create a variable for it */
```
young-k: In terms of maintaining this codebase for the future, the last sentence (beginning with `The extra 37...`) should be deleted, I don't think anyone will understand when you write `I felt it would be unnecessary...`

TCAT/Controllers/InformationViewController.swift
```diff
@@ -42,138 +43,137 @@ class InformationViewController: UIViewController, UITableViewDataSource, UITabl
...
+    private var headerHeight: CGFloat {
+        return tcatImageTopOffset + tcatImageSize.height + titleLabelTopOffset + titleLabelSize.height + descriptionLabelTopOffset + descriptionLabelSize.height + 37
+    }
+    private var tcatImageSize: CGSize { return CGSize(width: headerWidth-80, height: (headerWidth - 80) * 0.4) }
```
young-k: should be `headerWidth - 80` and not `headerWidth-80`, @kevinchan159 thoughts on making this multi-line? I'm personally not a fan of inlining `return` statements that have any sort of logic

TCAT/Controllers/InformationViewController.swift
```diff
@@ -42,138 +43,137 @@ class InformationViewController: UIViewController, UITableViewDataSource, UITabl
...
         dismissButton.addTarget(self, action: #selector(dismissTapped), for: .touchUpInside)
         dismissButton.setTitle(Constants.Buttons.done, for: .normal)
         navigationItem.rightBarButtonItem?.setTitleTextAttributes(
             CustomNavigationController.buttonTitleTextAttributes, for: .normal
         )
         let backButtonItem = UIBarButtonItem(customView: dismissButton)
         navigationItem.setRightBarButton(backButtonItem, animated: false)
+    }
 
+    func setupHiddenLabel() {
```
young-k: In terms of variable naming, it's really unclear to me why a "hidden label" would need to be set up –– is there a more descriptive variable name that would make this more clear?

TCAT/Controllers/InformationViewController.swift
```diff
@@ -42,138 +43,137 @@ class InformationViewController: UIViewController, UITableViewDataSource, UITabl
...
-        titleLabel.snp.makeConstraints { (make) in
-            make.top.equalTo(tcatImage.snp.bottom).offset(44)
-            make.centerX.equalToSuperview()
-            make.bottom.equalTo(descriptionLabel.snp.top).offset(-12)
-            make.height.equalTo(20)
-        }
+        headerView.addSubview(titleLabel)
+    }
 
+    func setupDescriptionLabel() {
```
young-k: nitpick, these functions should be alphabetized, which means that this should go above the others

TCAT/Controllers/InformationViewController.swift
```diff
@@ -351,3 +296,106 @@ class InformationViewController: UIViewController, UITableViewDataSource, UITabl
...
+        cell.selectionStyle = .none
+        cell.backgroundColor = Colors.white
+        cell.textLabel?.textColor = Colors.primaryText
+        cell.textLabel?.font = .getFont(.regular, size: 14)
+        cell.textLabel?.textAlignment = .center
+
+        // Set cell content
+        cell.textLabel?.text = content[indexPath.section][indexPath.row].name
+
+        // Set custom formatting based on cell
```
young-k, line 187: A `switch` statement seems like overkill here, since you're only checking one case?

TCAT/Controllers/InformationViewController.swift
```diff
@@ -351,3 +296,106 @@ class InformationViewController: UIViewController, UITableViewDataSource, UITabl
...
+extension InformationViewController: UITableViewDelegate {
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let selector = content[indexPath.section][indexPath.row].action
+        performSelector(onMainThread: selector, with: nil, waitUntilDone: false)
+    }
+}
+
+extension InformationViewController: MFMailComposeViewControllerDelegate {
+    @objc func sendFeedback() {
+
```
young-k, line 207: should there be a newline here? should there be a newline after declaring an `extension`? @kevinchan159 what are the rules

TCAT/Controllers/InformationViewController.swift
kevinchan159: yeah i think multiline is prob better

TCAT/Controllers/InformationViewController.swift
kevinchan159: i actually think it should be ordered in the order that they appear in the screen since thats how we build screens. alphabetizing it doesn't bring that much benefit especially since XCode has jump to definition

TCAT/Controllers/InformationViewController.swift
kevinchan159: we use `setup` pretty consistently over all codebases so i think we can keep it

TCAT/Controllers/InformationViewController.swift
kevinchan159: agree unless you plan on adding more cases in the near future

TCAT/Controllers/InformationViewController.swift
kevinchan159: yeah i pref newline at the beginning and end of `extension` bodies

TCAT/Controllers/InformationViewController.swift
kevinchan159: oh ok i see. +1 to young's comment

Issue #238, opened 2019-05-09
Refactor how we handle displaying bus stops in routeDetailVC


Omarrasheed:
Whenever we have a bus route that has several stops in between, we have a button to toggle whether or not to display the buses in that route. But the way we do this internally is by making an empty Direction object with a name and location and passing it into our list of directions. We should find a way to refactor this so that we can store and display both bus stops and directions so that we don't continue with this jank workaround

Omarrasheed:
Closing because this was fixed a while ago

PR #248, opened 2019-05-21
Refactor Route Detail VC

Omarrasheed:
#238 

Instead of storing bus stops as empty direction objects inserted into our vc's `self.directions` array, we now have a new enum `RouteDetailItem` which is either a BusStop (aka Location Object) or Direction, and our directions array is now of type `[RouteDetailItem]`

### DIFF
--- a/TCAT/Controllers/RouteDetail+ContentViewController.swift
+++ b/TCAT/Controllers/RouteDetail+ContentViewController.swift
@@ -15,12 +15,9 @@ import Pulley
 import SwiftyJSON
 import UIKit
 
-class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CLLocationManagerDelegate {
+class RouteDetailContentViewController: UIViewController {
 
     var drawerDisplayController: RouteDetailDrawerViewController?
-
-    private var mapView: GMSMapView!
-    private var routeOptionsCell: RouteTableViewCell?
 
     private var bounds = GMSCoordinateBounds()
     private var busIndicators = [GMSMarker]()
@@ -32,9 +29,11 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
     private var liveTrackingNetworkRefreshRate: Double = LiveIndicator.INTERVAL * 1.0
     private var liveTrackingNetworkTimer: Timer?
     private var locationManager = CLLocationManager()
+    private var mapView: GMSMapView!
     private let networking: Networking = URLSession.shared.request
     private var paths: [Path] = []
     private var route: Route!
+    private var routeOptionsCell: RouteTableViewCell?
 
     private var banner: StatusBarNotificationBanner? {
         didSet {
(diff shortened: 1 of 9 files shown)


young-k (approved):
LGTM, but you should probably wait for someone with more iOS knowledge

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -32,9 +29,11 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
     private var liveTrackingNetworkRefreshRate: Double = LiveIndicator.INTERVAL * 1.0
     private var liveTrackingNetworkTimer: Timer?
     private var locationManager = CLLocationManager()
+    private var mapView: GMSMapView!
     private let networking: Networking = URLSession.shared.request
     private var paths: [Path] = []
     private var route: Route!
+    private var routeOptionsCell: RouteTableViewCell?
```
kevinchan159, line 36: why do we need to store a reference to this?

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -741,4 +510,178 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
...
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        // Live Tracking Network Timer
+        liveTrackingNetworkTimer?.invalidate()
+        if directions.contains(where: { $0.type != .walk }) {
+            liveTrackingNetworkTimer = Timer.scheduledTimer(timeInterval: liveTrackingNetworkRefreshRate, target: self, selector: #selector(getBusLocations),
+                                                            userInfo: nil, repeats: true)
```
kevinchan159: if we're splitting this into two lines, we should have each arg on its own line
```
liveTrackingNetworkTimer = Timer.scheduledTimer(
  timeInterval: liveTrackingNetworkRefreshRate, 
  target: self, 
  selector: #selector(getBusLocations),
  userInfo: nil,
  repeats: true
)
```

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -741,4 +510,178 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
...
+    }
+
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        liveTrackingNetworkTimer?.invalidate()
+        hideBanner()
+    }
+
+    override func loadView() {
+
```
kevinchan159: delete newline

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -741,4 +510,178 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
...
+extension RouteDetailContentViewController: GMSMapViewDelegate {
+
+    func mapView(_ mapView: GMSMapView, didTap marker: GMSMarker) -> Bool {
+        guard let coordinates = getUserData(for: marker, key: Constants.BusUserData.actualCoordinates) as? CLLocationCoordinate2D
+            else { return true }
+        let update = GMSCameraUpdate.setTarget(coordinates)
+        mapView.animate(with: update)
+
+        return true
+
```
kevinchan159: delete newline

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -741,4 +510,178 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
...
+    }
+
+    func updateUserData(for marker: GMSMarker, with values: [String: Any]) {
+        guard var userData = marker.userData as? [String: Any] else {
+            marker.userData = values
+            return
+        }
+        for (key, value) in values {
```
kevinchan159: pref `forEach`

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -741,4 +510,178 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
...
+            userData[key] = value
+        }
+        marker.userData = userData
+    }
+
+    func getUserData(for marker: GMSMarker, key: String) -> Any? {
+        guard var userData = marker.userData as? [String: Any] else { return nil }
```
kevinchan159: does it have to be `var userData`? can it be `let`

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -741,4 +510,178 @@ class RouteDetailContentViewController: UIViewController, GMSMapViewDelegate, CL
...
+        return userData[key]
+    }
+
+    func mapView(_ mapView: GMSMapView, didChange position: GMSCameraPosition) {
+        for bus in buses {
```
kevinchan159: pref `forEach`

TCAT/Controllers/RouteDetail+DrawerViewController.swift
```diff
@@ -18,37 +18,52 @@ struct RouteDetailCellSize {
...
         super.init(nibName: nil, bundle: nil)
         self.route = route
-        self.directions = route.directions
+        self.directionsAndVisibleStops = route.directions.map({ RouteDetailItem.direction($0) })
     }
 
     func update(with route: Route) {
         self.route = route
-        self.directions = route.directions
+        self.directionsAndVisibleStops = route.directions.map({ RouteDetailItem.direction($0) })
```
kevinchan159, line 66: cleannn!

kevinchan159 (changes requested):
amazing!!

TCAT/Controllers/RouteDetail+ContentViewController.swift
Omarrasheed: It was being used to take a screenshot of the cell when a user shares but I'll make that a delegate function n just handle that in the main vc

Omarrasheed: Wait actually I'm not sure if we can do that. Cuz it takes an image of the route options cell which is like an overview of the route and then the user can share that specific route with someone else within the route details page. I'll think of a better way to do this but I'll probably leave this for now if that's ok

mattbarker016:
Reviewed 👍

PR #250, opened 2019-08-21
Omar/all stops table view controller

Omarrasheed:
Switches to normal VC + tableview instead of tableViewController

Will squash before merging


kevinchan159 (approved):
noice

PR #254, opened 2019-08-21
[ezpz] Make frame based layout more clear

Omarrasheed:
Seems like the underlying framework were using for the slides has its own constraint stuff thats being done under the hood and its being calculated from the frames. So I just cleaned up a tiny bit to make it more apparent that all we're setting is the sizes of the button/labels


kevinchan159 (approved):
lets get this money

PR #259, opened 2019-08-24
Use snapkit to layout routeDetailDrawerVC

Omarrasheed:
As title says. This tackles routeDetail+DrawerViewController and its extensions


young-k (approved):
:rocket:

PR #260, opened 2019-08-24
Remove frame based layout in circle

Omarrasheed:
As title says. Kinda reworked how we do the bordered circle (used for destinations)

### DIFF
--- a/TCAT/Views/Circle.swift
+++ b/TCAT/Views/Circle.swift
@@ -17,64 +17,62 @@ enum CircleStyle {
 }
 
 enum CircleSize: Int {
-    case large = 18
-    case medium = 16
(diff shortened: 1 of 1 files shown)


TCAT/Views/Circle.swift
```diff
@@ -26,7 +26,7 @@ class Circle: UIView {
 
     // MARK: Data vars
 
-    let diameter: CGFloat
+    private let diameter: CGFloat
```
young-k: Do we actually use the value of `diameter` directly anywhere, or is always `diameter / 2`? In the latter case, we should have a `radius` variable instead.

young-k (approved):

PR #263, opened 2019-08-25
Remove frame-based layout in liveIndicator

Omarrasheed:
Removed all frame-related code in here. There is no need to do autolayout on layers, as we manually create the paths of the layers with the UIBezierPath.

### DIFF
--- a/TCAT/Views/LiveIndicator.swift
+++ b/TCAT/Views/LiveIndicator.swift
@@ -7,7 +7,7 @@
 
 import UIKit
 
-enum LiveIndicatorSize: Double {
+enum LiveIndicatorSize: CGFloat {
     case large = 12
     case small = 8
 }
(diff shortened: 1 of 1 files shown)


TCAT/Views/LiveIndicator.swift
```diff
@@ -16,8 +16,10 @@ class LiveIndicator: UIView {
 
     // MARK: State vars
 
-    let lineWidth: CGFloat
-    let size: LiveIndicatorSize
+    private var lineWidth: CGFloat {
```
kevinchan159: does `size` ever change after we set its initial value in the `init`? if not can we just set `lineWidth` after we set `size`? typically computed properties are only used when its value may change over time

kevinchan159 (approved):
very noice
one question

TCAT/Views/LiveIndicator.swift
Omarrasheed: Updated!

PR #264, opened 2019-08-25
Use snapkit to layout busLocationView

Omarrasheed:
Seems like a LOT of the code we had in busLocationView was unused. 

Before deleting the `bearingIndicator`, I made sure that there was no code that was using or altering it. Before I had deleted, it was being initialized but set to hidden, and there was a function that could potentially unhide it but it was never being access.

Lastly, we still need to set the frame.size of the view because we use this within googleMaps' SDK as a marker and that I guess uses frame-based layout. So I removed as much frame-based layout as possible but technically not *all*

### DIFF
--- a/TCAT/Controllers/RouteDetail+ContentViewController.swift
+++ b/TCAT/Controllers/RouteDetail+ContentViewController.swift
@@ -73,7 +73,7 @@ class RouteDetailContentViewController: UIViewController {
         routeDetailViewController.navigationItem.setRightBarButton(shareButton, animated: true)
 
         // Debug Function
-        // createDebugBusIcon()
+//        createDebugBusIcon()
     }
 
     /** Construct Directions based on Route and parse Waypoint / Path data */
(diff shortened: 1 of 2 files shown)


TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -73,7 +73,7 @@ class RouteDetailContentViewController: UIViewController {
         routeDetailViewController.navigationItem.setRightBarButton(shareButton, animated: true)
 
         // Debug Function
-        // createDebugBusIcon()
+//        createDebugBusIcon()
```
young-k, line 76: is this indentation right?

young-k (approved):

kevinchan159 (approved):

TCAT/Controllers/RouteDetail+ContentViewController.swift
Omarrasheed: Yea it's j the default when u command + '/'.

Issue #252, opened 2019-08-21
Make unnecessary public functions into private functions


Omarrasheed:
As title says. This will cut down a lot on the number of things that are suggested in auto complete + ensures that other objects aren't able to use functions meant for internal use only

PR #269, opened 2019-08-28
Omar/private func refactor

Omarrasheed:
#252


TCAT/Controllers/HomeOptionsCardViewController.swift
```diff
@@ -320,9 +320,7 @@ class HomeOptionsCardViewController: UIViewController {
 // MARK: VC Life Cycle setup
 extension HomeOptionsCardViewController {
     override func loadView() {
-        let customView = RoundShadowedView()
-        customView.addRoundedCornersAndShadow()
-        view = customView
+        view = RoundShadowedView(cornerRadius: 10)
```
kevinchan159, line 323: do we need a separate function for this still?

kevinchan159 (approved):
lets get this $$$

TCAT/Controllers/HomeOptionsCardViewController.swift
Omarrasheed: Yea we still need this to override the main view before it loads

PR #271, opened 2019-08-29
Update to use new PlaceIDCoordinates route from backend

kevinchan159:
This PR:
- Now fetches the placeID coordinates from backend using the new `/placesIDCoordinate` route
- Fixes issue where `favorites` wasn't persisting between app launches
- Fixes app crashing when tapping on the `From` textfield in a search


young-k (approved):
approving to unblock

PR #273, opened 2019-08-29
Activate swiftlint for repo

Omarrasheed:


.swiftlint.yml
```diff
@@ -27,9 +30,6 @@ type_body_length:
 file_length:
 - 500
 
-line_length:
-- 120
```
young-k, line 31: Why are we getting rid of this? cc @kevinchan159

TCAT/Controllers/HomeOptionsCardViewController+Extensions.swift
```diff
@@ -0,0 +1,353 @@
...
+        switch section {
+        case .favorites:
+            if !section.isEmpty, section.getItems()[0].name != Constants.General.firstFavorite {
+                return .delete
+            } else {
+                return .none
+            }
+        case .recentSearches: return .delete
+        default: return .none
+        }
```
young-k, line 239: why not something like this:

```
if (section == .delete) || (section == .favorites && ...) {
  return .delete
} 
return .none
```

TCAT/Controllers/HomeOptionsCardViewController+Extensions.swift
```diff
@@ -0,0 +1,353 @@
...
+                Global.shared.insertPlace(for: Constants.UserDefaults.recentSearch, place: place)
+            }
+        }
+
+        tableView.deselectRow(at: indexPath, animated: true)
+        searchBar.endEditing(true)
+
+        let vcToPush = didSelectAllStops ? allStopsTableViewConroller : routeOptionsViewController
+        if shouldPushViewController {
+            navigationController?.pushViewController(vcToPush, animated: true)
```
kevinchan159, line 285: move 285 into the `if` block since we only use it if the condition is true

TCAT/Controllers/HomeOptionsCardViewController+Extensions.swift
```diff
@@ -0,0 +1,353 @@
...
+    func description(forEmptyDataSet scrollView: UIScrollView) -> NSAttributedString? {
+        // If loading indicator is being shown, don't display description
+        if isLoading { return nil }
+        let title = isNetworkDown ? Constants.EmptyStateMessages.noNetworkConnection: Constants.EmptyStateMessages.locationNotFound
+        return NSAttributedString(string: title, attributes: [.foregroundColor: Colors.metadataIcon])
+    }
+
+    func buttonTitle(forEmptyDataSet scrollView: UIScrollView, for state: UIControl.State) -> NSAttributedString? {
+        // If loading indicator is being shown, don't display button
+        if isLoading { return nil }
```
young-k: Is there no way to have all of these `isLoading` calls in one place?

TCAT/Controllers/OnboardingViewController.swift
```diff
@@ -163,55 +166,85 @@ class OnboardingViewController: PresentationController {
...
-        let width = UIScreen.main.bounds.width < 550 ? UIScreen.main.bounds.width : 550
-        let height: CGFloat = 200
+        let detailTitles = getDetailLabelTitles()
+        let headerTitles = getHeaderLabelTitles()
+        let startButton = Content(view: getBeginButton(), position: Position(left: 0.5, top: 0.5), centered: true)
+
+        // Slides
+        var slides = [SlideController]()
+
+        for index in 0..<detailLabelMessages.count {
```
kevinchan159: I prefer `(0..<detailLabelMessages.count).forEach()`

TCAT/Controllers/RouteDetailContentViewController+Extensions.swift
```diff
@@ -0,0 +1,191 @@
...
+                        Constants.BusUserData.indicatorCoordinates: placement
+                        ])
+
+                    existingIndicator.appearAnimation = .none
+                    // existingIndicator.map = mapView // Uncomment to avoid animation
+                } else {
+                    existingIndicator.map = nil
+                    busIndicators.remove(at: busIndicators.firstIndex(of: existingIndicator)!)
+                }
+            } else { // Create Indicator
```
kevinchan159: might be easier to read if you add a return after line 150 and then remove this `else` block

TCAT/Model/Route.swift
```diff
@@ -122,86 +122,83 @@ class Route: NSObject, Codable {
...
-            // Change all walking directions, except for first and last direction, to arrive
-            let last = rawDirections.count - 1
-            for (index, direction) in rawDirections.enumerated() {
-                if index != last && index != first && direction.type == .walk {
-                    direction.type = .arrive
-                    direction.name = rawDirections[index - 1].stops.last?.name ?? "Bus Stop"
-                }
+        // Change all walking directions, except for first and last direction, to arrive
+        let last = rawDirections.count - 1
+        for (index, direction) in rawDirections.enumerated() {
```
kevinchan159: pref `forEach`

TCAT/Model/Route.swift
```diff
@@ -122,86 +122,83 @@ class Route: NSObject, Codable {
...
-                    direction.name = startName
-                } else {
-                    direction.name = rawDirections[index - 1].stops.last?.name ?? rawDirections[index - 1].name
-                }
+    func formatDirections(start: String?, end: String?) {
+        startName = start ?? Constants.General.currentLocation
+        endName = end ?? Constants.General.destination
+
+        let first = 0
+        for (index, direction) in rawDirections.enumerated() where direction.type == .walk {
```
kevinchan159, line 130: would be more functional to use a `filter` and `forEach` but you can keep this if its more readable

.swiftlint.yml
kevinchan159: i spoke w/ @Omarrasheed about it and i think its b/c right now there are hundreds of warnings for this rule if we activated it. I think we should make a task to fix all the warning and add it sometime in the future @Omarrasheed

young-k: Why not just set it really high, and lower it in the future?

kevinchan159: we could do that too. @Omarrasheed try to find the smallest number you can use

TCAT/Controllers/HomeOptionsCardViewController+Extensions.swift
Omarrasheed: I can't directly use '==' with just the section types because the enums have associated values; I'd need to create an actual Section.recentSearches with an associated value and *then* compare it, but at that point it would be comparing the values too and not just the types. Could write a function called .isEqualType(to: ) but idk if that really helps much. Opinions? cc @kevinchan159

TCAT/Controllers/HomeOptionsCardViewController+Extensions.swift
Omarrasheed: Will remove it completely. I noticed we were never even setting the delegates to self and thus were never even using all of the loading indicator stuff.

TCAT/Controllers/HomeOptionsCardViewController+Extensions.swift
kevinchan159: i think you can do `if (case .delete == section) || (case .favorites == section && ...) { }`

Omarrasheed: I can't get it to work :(

young-k (approved):

PR #275, opened 2019-08-30
[Very easy] Stop using current location to show route

Omarrasheed:
Idk y we were using current location to show the route overview but I took it out because it was causing problems especially when the user is out of state. Furthermore, I found that for cases where the user isn't using current location and simply searching from one place to another, including the current location in the zoomed in map view didn't seem to be very useful.

### New
Screen Shot 2019-08-29 at 10 56 49 PM

### Old
Screen Shot 2019-08-29 at 10 56 26 PM


kevinchan159 (approved):
can you take a screenshot/video of how it looks like now?

Omarrasheed:
@kevinchan159 Updated! New is what it looks like whether or not the user is in the location range

kevinchan159:
lgtm

Issue #211, opened 2019-03-21
Max width iOS issue


mayafrai:
Routes with more than 10 stops are not visible on the bottom sheet when expanded. 

Investigate the max-width set for the bottom sheet.

Example: 
![IMG_0446](https://user-images.githubusercontent.com/25111415/54721327-112e2780-4b38-11e9-9048-53ffe58ef5c7.PNG)

@ywsang

Omarrasheed:
Closing because fixed with #276

PR #276, opened 2019-08-30
Add content inset to account for safe area

Omarrasheed:
Knocks out #211

### New:
Screen Shot 2019-08-29 at 11 44 00 PM

### Old:
Screen Shot 2019-08-29 at 11 49 32 PM


kevinchan159 (approved):
lgtm

PR #277, opened 2019-08-30
Move debug scheme to new bundle ID

ywsang:
Moves the "Debug" scheme to a new bundle ID, `com.cornellappdev.tcat.debug` so that a version pointing to the dev server and a version pointing to the prod server can be installed on one device at the same time.

*will put all the new and/updated plist files on slack tomorrow* so this won't work until you update those files locally


Omarrasheed (approved):
A TANK!

TCAT/Supporting Files/Keys.swift
```diff
@@ -13,8 +13,10 @@ enum Keys: String {
     case fabricAPIKey = "fabric-api-key"
     case fabricBuildSecret = "fabric-build-secret"
 
-    case googleMaps = "google-maps"
-    case googlePlaces = "google-places"
+    case googleMapsDebug = "google-maps-debug"
+    case googleMapsRelease = "google-maps-release"
+    case googlePlacesDebug = "google-places-debug"
+    case googlePlacesRelease = "google-places-release"
```
kevinchan159: I believe I deleted `GooglePlaces` entirely from the PR i made yesterday so I think you can just delete these two keys

kevinchan159 (approved):
amazing! just 1 comment

Issue #279, opened 2019-09-01
Text label is being cut off


Omarrasheed:
### Bug:
<img src="https://user-images.githubusercontent.com/36868927/64079103-59d63900-ccb1-11e9-8859-9e7476365022.jpg" width="300">

lucyxubroad:
Resolved in  PR #288

PR #288, opened 2019-09-05
Fix stopLabel width and word wrap.

lucyxubroad:
- Addresses Issue #279 Text label is being cut off
- Bug screen with text label cut off: 

- Fixed screen with text wrap: 
Screen Shot 2019-09-05 at 4 58 37 PM Screen Shot 2019-09-05 at 4 53 38 PM

- Changes: Auto layout may have been interfering with stopLabel frame, preventing text wrap. I removed the frame and moved the width constraint into auto layout. getTestStopLabel(name) was also incorrectly calculating the testStopLabel width, causing overflow.

### DIFF
--- a/TCAT/Views/RouteDiagramSegment.swift
+++ b/TCAT/Views/RouteDiagramSegment.swift
@@ -82,6 +82,7 @@ class RouteDiagramSegment: UIView {
 
         stopLabel.snp.makeConstraints { make in
             make.leading.equalTo(stopDot.snp.centerX).offset(spaceBtnStopDotCenterXAndStopLabel)
+            make.trailing.equalToSuperview()
             make.top.equalTo(stopDot)
         }
     }
@@ -89,11 +90,12 @@ class RouteDiagramSegment: UIView {
     // MARK: Get data from route ojbect
 
     private func getStopLabel(withName name: String, withStayOnBusForTranfer stayOnBusForTranfer: Bool, withDistance distance: Double?) -> UILabel {
-        let yPos: CGFloat = 101
-        let rightSpaceFromSuperview: CGFloat = 16
-        let width: CGFloat = UIScreen.main.bounds.width - yPos - rightSpaceFromSuperview
+        let labelPadding: CGFloat = 12
+        let rightPadding: CGFloat = 16
+        let xPos: CGFloat = 101
+        let width: CGFloat = UIScreen.main.bounds.width - xPos - rightPadding - labelPadding
 
-        let stopLabel = UILabel(frame: CGRect(x: 0, y: 0, width: width, height: 17))
+        let stopLabel = UILabel()
         // allow for multi-line label for long stop names
         stopLabel.allowsDefaultTighteningForTruncation = true
         stopLabel.lineBreakMode = .byWordWrapping
@@ -143,7 +145,7 @@ class RouteDiagramSegment: UIView {
         testStopLabel.font = .getFont(.regular, size: 14.0)
         testStopLabel.textColor = Colors.primaryText
         testStopLabel.text = name
-
+        testStopLabel.sizeToFit()
         return testStopLabel
     }
 


TCAT/Views/RouteDiagramSegment.swift
```diff
@@ -79,9 +82,10 @@ class RouteDiagramSegment: UIView {
             make.centerX.equalTo(snp.leading).offset(circleCenterXLeadingInset)
             make.size.equalTo(stopDot.intrinsicContentSize)
         }
-
+
```
Omarrasheed: nit: delete

TCAT/Views/RouteDiagramSegment.swift
```diff
@@ -143,7 +147,7 @@ class RouteDiagramSegment: UIView {
         testStopLabel.font = .getFont(.regular, size: 14.0)
         testStopLabel.textColor = Colors.primaryText
         testStopLabel.text = name
-
+        testStopLabel.sizeToFit()
```
Omarrasheed, line 148: We shouldn't need this

TCAT/Views/RouteDiagramSegment.swift
```diff
@@ -51,7 +51,10 @@ class RouteDiagramSegment: UIView {
             return self.icon is UIImageView ? spaceBtnWalkIconAndSuperview : (self.icon is BusIcon ? 0 : 8.5)
         }
         let spaceBtnStopDotCenterXAndStopLabel: CGFloat = 26
-
+        let screenWidth = UIScreen.main.bounds.width
+        let stopLabelRightPadding: CGFloat = 16
+        let stopLabelYPosition: CGFloat = 101
```
Omarrasheed: is this X or Y position?

TCAT/Views/RouteDiagramSegment.swift
```diff
@@ -93,7 +97,7 @@ class RouteDiagramSegment: UIView {
         let rightSpaceFromSuperview: CGFloat = 16
         let width: CGFloat = UIScreen.main.bounds.width - yPos - rightSpaceFromSuperview
 
-        let stopLabel = UILabel(frame: CGRect(x: 0, y: 0, width: width, height: 17))
+        let stopLabel = UILabel()
```
Omarrasheed, line 98: Good catch

TCAT/Views/RouteDiagramSegment.swift
```diff
@@ -79,9 +82,10 @@ class RouteDiagramSegment: UIView {
             make.centerX.equalTo(snp.leading).offset(circleCenterXLeadingInset)
             make.size.equalTo(stopDot.intrinsicContentSize)
         }
-
+        
         stopLabel.snp.makeConstraints { make in
             make.leading.equalTo(stopDot.snp.centerX).offset(spaceBtnStopDotCenterXAndStopLabel)
+            make.width.equalTo(stopLabelWidth)
```
Omarrasheed: Can we look into setting a trailing constraint rather than calculating a width?

Omarrasheed (changes requested):

Omarrasheed:
Forgot to comment but screenshots look good. Would like to see if we can j set a trailing constraint instead of calculating the width dynamically

TCAT/Views/RouteDiagramSegment.swift
Omarrasheed: JK seems likes we may need this cuz of the jank stuff we do in setting the label text using some fake test label. If u waaaaant u could look into it but not necessary for now

TCAT/Views/RouteDiagramSegment.swift
```diff
@@ -51,7 +51,10 @@ class RouteDiagramSegment: UIView {
             return self.icon is UIImageView ? spaceBtnWalkIconAndSuperview : (self.icon is BusIcon ? 0 : 8.5)
         }
         let spaceBtnStopDotCenterXAndStopLabel: CGFloat = 26
-
+        let screenWidth = UIScreen.main.bounds.width
+        let stopLabelRightPadding: CGFloat = 16
+        let stopLabelYPosition: CGFloat = 101
+        let stopLabelWidth: CGFloat = screenWidth - stopLabelYPosition - stopLabelRightPadding
```
kevinchan159: should we be subtracting `stopLabelYPosition` or `stopLabelXPosition` since its width?

kevinchan159 (commented):
nice work!!! 🤗
had similar questions as @Omarrasheed

TCAT/Views/RouteDiagramSegment.swift
lucyxubroad: Maybe I'll look into this later!

lucyxubroad:
Made the changes and the code is much simpler! I also added some padding to the width calculation because I think there are some inconsistencies between the width that Swift calculates for labels and the labels we have on Zeplin. Adding the extra padding ensures our distance label never gets cut off.

TCAT/Views/RouteDiagramSegment.swift
```diff
@@ -82,18 +81,20 @@ class RouteDiagramSegment: UIView {
 
         stopLabel.snp.makeConstraints { make in
             make.leading.equalTo(stopDot.snp.centerX).offset(spaceBtnStopDotCenterXAndStopLabel)
+            make.trailing.equalToSuperview()
```
Omarrasheed, line 85: Can we just add an inset? 16 should be good

Omarrasheed (approved):
Looks good! Just add the trailing inset before shipping :) Also go to bed lmao

TCAT/Views/RouteDiagramSegment.swift
```diff
@@ -82,18 +81,20 @@ class RouteDiagramSegment: UIView {
...
     // MARK: Get data from route ojbect
 
     private func getStopLabel(withName name: String, withStayOnBusForTranfer stayOnBusForTranfer: Bool, withDistance distance: Double?) -> UILabel {
-        let yPos: CGFloat = 101
-        let rightSpaceFromSuperview: CGFloat = 16
-        let width: CGFloat = UIScreen.main.bounds.width - yPos - rightSpaceFromSuperview
+        let xPos: CGFloat = 101
+        let rightPadding: CGFloat = 16
+        let labelPadding: CGFloat = 12
+        let width: CGFloat = UIScreen.main.bounds.width - xPos - rightPadding - labelPadding
```
YKo20010, line 96: alphabetize

YKo20010 (approved):
Just one alphabetize comment, but lgtm! :)

Issue #280, opened 2019-09-01
Update "leave now" time when refreshing routes


Omarrasheed:
When a user leaves their app and comes back, we seem to update the routes properly but the time text that goes along with "Leave Now" doesn't seem to get updated

### Bug:
<img src="https://user-images.githubusercontent.com/36868927/64079222-28f70380-ccb3-11e9-989c-0d6f3b8261d8.png" width="300">

PR #290, opened 2019-09-06
Refresh routes and leave times

usubzero:
Refreshes routes and leave now/at/by time whenever the app is reopened or the time has changed while the user was looking at a route detail

Addresses issue #280  and more

### DIFF
--- a/TCAT/Controllers/RouteOptionsViewController.swift
+++ b/TCAT/Controllers/RouteOptionsViewController.swift
@@ -91,7 +91,9 @@ class RouteOptionsViewController: UIViewController {
 
         // assume user wants to find routes that leave at current time and set datepicker accordingly
         searchTime = Date()
-        routeSelection.setDatepickerTitle(withDate: searchTime!, withSearchTimeType: searchTimeType)
+        if let searchTime = searchTime {
+            routeSelection.setDatepickerTitle(withDate: searchTime, withSearchTimeType: searchTimeType)
+        }
 
         searchForRoutes()
 
@@ -106,7 +108,6 @@ class RouteOptionsViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         setupReachability()
-        
         // Reload data to activate timers again
         if !routes.isEmpty {
             routeResults.reloadData()
@@ -330,12 +331,10 @@ class RouteOptionsViewController: UIViewController {
     
     @objc private func refreshRoutesAndTime() {
         let now = Date()
-        lastRouteRefreshDate = now
-        if searchTimeType != .leaveNow {
-            if let leaveDate = searchTime,
-                leaveDate.compare(now) == .orderedDescending {
-                return
-            }
+        if let leaveDate = searchTime,
+            searchTimeType != .leaveNow,
+            leaveDate.compare(now) == .orderedDescending {
+            return
         }
         
         searchTime = now
@@ -357,6 +356,8 @@ class RouteOptionsViewController: UIViewController {
         if let searchFrom = searchFrom,
             let searchTo = searchTo,
             let time = searchTime {
+            let now = Date()
+            lastRouteRefreshDate = now
             resetAndShowCurrentlyLoading()
 
             switch searchType {
@@ -387,7 +388,7 @@ class RouteOptionsViewController: UIViewController {
             // Prepare feedback on Network request
             mediumTapticGenerator.prepare()
 
-            JSONFileManager.shared.logSearchParameters(timestamp: Date(), startPlace: searchFrom, endPlace: searchTo, searchTime: time, searchTimeType: searchTimeType)
+            JSONFileManager.shared.logSearchParameters(timestamp: now, startPlace: searchFrom, endPlace: searchTo, searchTime: time, searchTimeType: searchTimeType)
 
             // MARK: Search For Routes Errors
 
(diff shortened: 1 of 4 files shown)


TCAT/Views/RouteSelectionView.swift
```diff
@@ -18,7 +18,7 @@ protocol RouteSelectionViewDelegate: class {
 class RouteSelectionView: UIView {
 
     private weak var delegate: RouteSelectionViewDelegate?
-
+
```
Omarrasheed, line 21: nit: delete

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -107,11 +106,13 @@ class RouteOptionsViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         setupReachability()
-
+
```
Omarrasheed: del

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -90,9 +91,7 @@ class RouteOptionsViewController: UIViewController {
 
         // assume user wants to find routes that leave at current time and set datepicker accordingly
         searchTime = Date()
-        if let searchTime = searchTime {
-            routeSelection.setDatepickerTitle(withDate: searchTime, withSearchTimeType: searchTimeType)
-        }
+        routeSelection.setDatepickerTitle(withDate: searchTime!, withSearchTimeType: searchTimeType)
```
Omarrasheed: y r we force unwrapping?

TCAT/Controllers/RouteOptionsViewController.swift
kevinchan159: delete whitespace

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -311,6 +327,22 @@ class RouteOptionsViewController: UIViewController {
             }
         }
     }
+    
+    @objc private func refreshRoutesAndTime() {
+        let now = Date()
+        lastRouteRefreshDate = now
+        if searchTimeType != .leaveNow {
+            if let leaveDate = searchTime,
```
Omarrasheed: can put these two in the same line. So 
```swift
if searchTimeType != .leaveNow,
  let leaveDate = searchTime {
  ...
}
```

TCAT/Controllers/RouteOptionsViewController.swift
kevinchan159: can we unwrap `searchTime`. I know in this case searchTime isn't going to be nil but its safer to unwrap in case someone ends up changing this code in the future

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -311,6 +327,22 @@ class RouteOptionsViewController: UIViewController {
             }
         }
     }
+    
+    @objc private func refreshRoutesAndTime() {
+        let now = Date()
+        lastRouteRefreshDate = now
```
Omarrasheed: would move this down into the `searchForRoutes` to make sure it's updated *every* time we call `searchForRoutes`

Omarrasheed (changes requested):
Looks great! Some minor nits + make sure we're updating `lastRouteRefreshDate` everytime we refresh.

TCAT/Views/RouteSelectionView.swift
kevinchan159, line 21: delete whitespace

usubzero:
Moved update to lastRouteRefreshDate to be within searchForRoutes() rather than in refreshRoutesAndTime()

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -47,6 +47,7 @@ class RouteOptionsViewController: UIViewController {
     var searchTimeType: SearchType = .leaveNow
     var searchTo: Place?
     var searchType: SearchBarType = .to
+    var lastRouteRefreshDate = Date()
     var showRouteSearchingLoader: Bool = false
```
YKo20010: alphabetize

YKo20010 (approved):

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -194,6 +198,19 @@ class RouteOptionsViewController: UIViewController {
...
         hideSearchBar()
     }
+    
+    private func setUpRouteRefreshing() {
+        let now = Date()
+        let hourMinuteComponents: Set<Calendar.Component> = [.hour, .minute]
+        let nowTime = Calendar.current.dateComponents(hourMinuteComponents, from: now)
+        let lastRefreshTime = Calendar.current.dateComponents(hourMinuteComponents, from: lastRouteRefreshDate)
+        if nowTime != lastRefreshTime {
+            refreshRoutesAndTime()
```
Omarrasheed, line 208: Just curious, y do we have to have this call in the initialization?

Omarrasheed (approved):
Looks good fam. Just had one more question

TCAT/Views/RouteSelectionView.swift
kevinchan159, line 21: theres still whitespace here :O

TCAT/Controllers/RouteOptionsViewController.swift
usubzero: It's called in viewWillAppear so that when somebody goes back to this VC from the route detail VC it only refreshes time and routes if the time changed. I chose to not include this bit of code within refreshRoutesAndTime so that the function produces the expected results in any call regardless of time.

TCAT/Views/RouteSelectionView.swift
usubzero: All throughout the app there's whitespace between delegate declarations and other variable declarations. This white space was there since before I did any edits (I had inserted a variable and removed it so Github marked an inexistent change; hence the +- whitespace)

Issue #292, opened 2019-09-08
Make sure to configure datePickerView given previously selected options


Omarrasheed:
Make sure that settings selected (arrive by and time) and being passed in and out of the View properly. Especially when making the request too

PR #298, opened 2019-09-09
Appropriately saves date picker search type state

usubzero:
Updated the doneButtonPressed method in DatePickerView to fix the assignment to searchTimeType variable. Also updated the saveDatePickerDate(Date, SearchType) method in RouteOptionsViewController to store this passed state (passed from doneButtonPressed) in RoutesOptionsViewController.

### DIFF
--- a/TCAT/Controllers/RouteOptionsViewController+Extensions.swift
+++ b/TCAT/Controllers/RouteOptionsViewController+Extensions.swift
@@ -101,9 +101,9 @@ extension RouteOptionsViewController: DatePickerViewDelegate {
         routeSelection.setDatepickerTitle(withDate: date, withSearchTimeType: searchTimeType)
         var buttonTapped = ""
         switch searchType {
-        case .leaveNow: buttonTapped = "Leave Now Tapped"
         case .arriveBy: buttonTapped = "Arrive By Tapped"
         case .leaveAt: buttonTapped = "Leave At Tapped"
+        case .leaveNow: buttonTapped = "Leave Now Tapped"
         }
         dismissDatePicker()
 
--- a/TCAT/Controllers/RouteOptionsViewController+Extensions.swift
+++ b/TCAT/Controllers/RouteOptionsViewController+Extensions.swift
@@ -96,24 +96,21 @@ extension RouteOptionsViewController: DatePickerViewDelegate {
 
     func saveDatePickerDate(for date: Date, searchType: SearchType) {
         searchTime = date
+        searchTimeType = searchType
 
         routeSelection.setDatepickerTitle(withDate: date, withSearchTimeType: searchTimeType)
-
         var buttonTapped = ""
-
         switch searchType {
         case .leaveNow: buttonTapped = "Leave Now Tapped"
         case .arriveBy: buttonTapped = "Arrive By Tapped"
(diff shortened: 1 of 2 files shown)


usubzero:
Refers to issue #292

TCAT/Views/DatePickerView.swift
```diff
@@ -183,13 +183,14 @@ class DatePickerView: UIView {
...
             searchTimeType = .leaveNow
-        case arriveByElement.index:
-            searchTimeType = .arriveBy
-        default:
+        } else if timeTypeSegmentedControl.selectedSegmentIndex == leaveAtElement.index {
             searchTimeType = .leaveAt
+        } else if timeTypeSegmentedControl.selectedSegmentIndex == arriveByElement.index {
+            searchTimeType = .arriveBy
+        } else {
+            searchTimeType = .leaveNow
```
Omarrasheed: Can we keep this as a switch rather than a bunch of else ifs?

Omarrasheed (approved):
LGTM. Lets just either keep the switch or give a good reason to move off of it

TCAT/Views/DatePickerView.swift
```diff
@@ -183,13 +183,14 @@ class DatePickerView: UIView {
...
+        if leaveNowSegmentedControl.selectedSegmentIndex == leaveNowElement.index {
             searchTimeType = .leaveNow
-        case arriveByElement.index:
-            searchTimeType = .arriveBy
-        default:
+        } else if timeTypeSegmentedControl.selectedSegmentIndex == leaveAtElement.index {
             searchTimeType = .leaveAt
+        } else if timeTypeSegmentedControl.selectedSegmentIndex == arriveByElement.index {
+            searchTimeType = .arriveBy
+        } else {
```
kevinchan159: prefer to use a `switch` for enums

kevinchan159 (changes requested):
nice job! 😁 just one question. 
also please update the PR description to also include a short description of how you solved the issue/screenshots!

TCAT/Controllers/RouteOptionsViewController+Extensions.swift
```diff
@@ -96,24 +96,21 @@ extension RouteOptionsViewController: DatePickerViewDelegate {
...
+        searchTimeType = searchType
 
         routeSelection.setDatepickerTitle(withDate: date, withSearchTimeType: searchTimeType)
-
         var buttonTapped = ""
-
         switch searchType {
         case .leaveNow: buttonTapped = "Leave Now Tapped"
         case .arriveBy: buttonTapped = "Arrive By Tapped"
         case .leaveAt: buttonTapped = "Leave At Tapped"
```
young-k, line 105: prefer that these cases are alphabetized

TCAT/Views/DatePickerView.swift
usubzero: Incorporated an improvement in logic that makes the switch more suitable. I initially had it all as else if's since the first comparison was of a different variable, so thought it'd look better for consistency. Update coming

usubzero:
Alphabetized switch and improved logic for determining searchTimeType

Omarrasheed (approved):
LGTM

kevinchan159 (approved):

PR #301, opened 2019-09-20
Add Local build configuration

kevinchan159:
As title says, this PR adds a new Local build configuration which allows us to then create a new `TCAT Local` build scheme. This build scheme uses the `Local` build configuration like so:
Screen Shot 2019-09-20 at 1 12 07 AM

Then, inside `Endpoints.swift`, we can use the `#if LOCAL` check to use `http` and port `3000`. Also notice that `SERVER_URL = localhost;` in the `project.pbxproj` for the `Local` build config.

As a result, now we can run `TCAT Local` to query the local backend server. This is useful for when we want to debug locally and add print statements in our local backend

### DIFF
--- a/TCAT.xcodeproj/project.pbxproj
+++ b/TCAT.xcodeproj/project.pbxproj
@@ -199,8 +199,10 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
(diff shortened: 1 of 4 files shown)


Omarrasheed (approved):
Bet

young-k (approved):

lucyxubroad (approved):
lgtm

PR #303, opened 2019-09-24
Convert from GoogleMaps to Apple's MapKit

kevinchan159:
After making the necessary backend changes (in PR: https://github.com/cuappdev/ithaca-transit-backend/pull/260), I pushed an image to the dev server. I tested those changes along with these changes and have recorded a demo video demonstrating that searching still works.

**NOTE: I have posted this demo video in `#transit-ios`.**

As a summary, we currently have 2 main uses for GooglePlaces:
1. Autocomplete: When a user types a query, i.e. `Duff`, we query GooglePlaces to get a list of places to display, such as `Duffield Hall`. Each `googlePlace` here has an associated `placeIdentifier`.
2. PlacesDetails: This is used to convert a `googlePlace`'s `placeIdentifier` to its corresponding `lat` and `long` coordinates.

To replace this completely, I made two new routes on the backend. The new flow for searching is as follows:

When a user makes a search on iOS for a query, i.e. `Duff`, client will hit `/appleSearch` which will return an array of busStops AND an array of applePlaces, if those places exist in the backend cache. If it doesn't, we just return an array of `busStops`. On client-side, when we see that server did not return any `applePlaces`, we will do the lookup for ApplePlaces on iOS and then send that array of `applePlaces` to `/applePlaces` which will just store these places in the cache. Now, the next time a user types in `Duff`, server will just return both the busStops and applePlaces and client will not need to do the lookup for ApplePlaces.

Some specific changes in this PR:
- Remove `CoordinateVisitor` which was previously being used to query the backend to convert `placeIdentifier` to `lat` and `long`
- Added networking for `/appleSearch` and `/applePlaces`
- Created a singleton `SearchManager` which uses Apple's MapKit to perform autocomplete and also acquires `lat/long` for all searched places

I also tested that `applePlaces` are being cached on backend:
Screen Shot 2019-09-23 at 10 18 22 PM


TCAT/Controllers/FavoritesTableViewController.swift
```diff
@@ -180,21 +155,15 @@ extension FavoritesTableViewController: UISearchBarDelegate {
...
-                        if response.success {
-                            self.resultsSection = Section.searchResults(items: response.data)
-                        } else {
-                            print("[FavoritesTableViewController] success: false")
-                            self.resultsSection = Section.searchResults(items: [])
-                        }
-                    case .error(let error):
-                        print("[FavoritesTableViewController] getSearchResults Error: \(error.localizedDescription)")
+                    if let error = error {
+                        print("[FavoritesTableViewController] SearchManager lookup Error: \(error.localizedDescription)")
```
Omarrasheed, line 162: Can we make an issue to standardize handling and logging of errors

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -373,24 +352,6 @@ class RouteOptionsViewController: UIViewController {
                 routeSelection.updateSearchBarTitles(to: searchTo.name)
             }
 
-            // If don't have coordinates, fetch and restart process
-            if searchFrom.latitude == nil || searchFrom.longitude == nil {
```
Omarrasheed, line 377: Are we sure we can take this out? Idk the answer but j checking

Omarrasheed (approved):
LGTM
~Can u give me, Lucy, n Haiying a quick rundown on this stuff + caching on wednesday or sunday?~

SIKE just read ur summary lmao

TCAT/Controllers/RouteOptionsViewController.swift
kevinchan159: Yeah I'm pretty confident we can. We had this in the past because when we had a `googlePlace`, it was in either one of two states: 1) we had its `placeIdentifier` but not its coordinates or 2) we had its `placeIdentifier` and its coordinates (which we obtained through `GooglePlacesDetails`. For MapKit, all `applePlaces` should be created w/ `lat` and `long` already and there isn't this two step process anymore. In terms of `busStops`, backend handles providing all coordinates for them

young-k (approved):

PR #304, opened 2019-09-24
Disable dark mode

kevinchan159:
App looks a bit more normal now 😂
Screen Shot 2019-09-23 at 10 39 33 PM
Screen Shot 2019-09-23 at 10 39 38 PM


Omarrasheed (approved):
Bet LGTM

PR #307, opened 2019-09-25
Handle decode error when delay key is missing

kevinchan159:
Currently, there is a bug where for some searches, even though the backend returns the response, we get a `Could not connect to server` error. After some investigation, I realized its because there was a `Decoding` error where a `Direction` object was sometimes missing the `delay` key` ,as in the following screenshot:

Screen Shot 2019-09-24 at 11 15 04 PM
**Notice how the object in index 0 has a `delay` key but the object in index 1 does not**

To fix this, I wrapped the decoding in a `do {} catch {}`.

Before:
Screen Shot 2019-09-24 at 11 11 48 PM

After:
Screen Shot 2019-09-24 at 11 13 02 PM


alanna-zhou (approved):
go go go

PR #309, opened 2019-09-26
iOS13 Fixes before pushing

kevinchan159:
This PR fixes some of the more noticeable bugs that appear

- Fix the searchIcon on home screen from becoming extremely big. 
**FIX**: In `RouteOptionsViewController`, we create a `SearchBarView` and for some reason in the `SearchBarView:init`, we were setting the image of UISearchBar globally like so: `UISearchBar.appearance().image...`. I removed this code and now the bug is gone
- In `RouteOptionsViewController`, when you tap on the `From` or `To` textfield and make a search, the top of the tableView from `SearchResultsTableViewController` is hidden behind the searchbar in the navigation bar.
**FIX**: Convert `SearchResultsTableViewController` into a `UIViewController` with a `UITableView` and set constraints on the table view
- After changing the `From/To` textfield and then selecting a new location, the navigation bar height isn't restored so there is a huge gap right below the navigation bar
**FIX**: Call `setNeedsLayout()` on the `navigationBar.view` after we set the `navigationItem.searchController` so that the `navigationBar`'s height get adjusted appropriately

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -16,6 +16,7 @@ target 'TCAT' do
     pod 'SwiftyJSON', '~> 5.0'
     pod 'FutureNova', :git => 'https://github.com/cuappdev/ios-networking.git'
     pod 'Wormholy', :configurations => ['Debug']
+    pod 'FLEX', :configurations => ['Debug']
     
     # Analytics
     pod 'Crashlytics', '~> 3.12'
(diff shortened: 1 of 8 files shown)


Omarrasheed (approved):
SHIP

Issue #289, opened 2019-09-05
Get rid of test label functions in routeDiagramSegment


Omarrasheed:
We currently have two functions `getTestStopLabel` and `getTestDistanceLabel` that we use for some black magic text and size calculations. We should definitely get rid of this cuz it doesn't make sense that we need a fake label to calculate information about our real label.

PR #312, opened 2019-09-29
Lucy/stop label changes

lucyxubroad:
Addresses #289 

Initially, PR was to remove the test label functions to compute the width of "test labels" before we add the actual label onto the screen, which was jank and not ideal. After some playing around, I think that we actually need to have the function because without it, we cannot manually add a line break between our stop name and stop distance. 
- So instead of getting rid of the functions completely, I just consolidated the two into a more simplified function that takes into account both stop name and stop distance but doesn't account for miscellaneous style stuff. 
- I also added preferredMaxWidth to make sure text overflow happens
- I fixed the constants we're currently using for max width calculation for our labels

Note: Currently, the app crashes because of a mix of iOS things and live updates so we might wanna hold off on merging this PR?

Screen Shot 2019-09-29 at 1 26 17 PM


kevinchan159 (approved):
amazing!

Omarrasheed (approved):
Goat tyty

Issue #313, opened 2019-09-29
Fix all stops sorting


haiyingweng:
Some stops that should be in # are in A
![Simulator Screen Shot - iPhone Xʀ - 2019-09-29 at 13 54 05](https://user-images.githubusercontent.com/48968041/65836943-c8aeae00-e2c0-11e9-8e4c-df52685f8b6b.png)

PR #314, opened 2019-10-01
Fix all stops sorting

haiyingweng:
For #313 

All stops were sorted incorrectly. Some stops that begin with a number are sorted into A. This PR fixes the problem. 

Before: 

After:

### DIFF
--- a/TCAT/Controllers/AllStopsTableViewController.swift
+++ b/TCAT/Controllers/AllStopsTableViewController.swift
@@ -67,19 +67,17 @@ class AllStopsTableViewController: UIViewController {
         var sectionIndexDictionary: [String: [Place]] = [:]
         var numberBusStops: [Place] = []
 
-        if !allStops.isEmpty {
-            allStops.forEach { busStop in
-                if let firstChar = busStop.name.capitalized.first,
-                    let firstScalar = firstChar.unicodeScalars.first {
-                    if CharacterSet.decimalDigits.contains(firstScalar) {
-                        numberBusStops.append(busStop)
+        allStops.forEach { busStop in
+            if let firstChar = busStop.name.capitalized.first,
+                let firstScalar = firstChar.unicodeScalars.first {
+                if CharacterSet.decimalDigits.contains(firstScalar) {
+                    numberBusStops.append(busStop)
+                } else {
+                    if var stops = sectionIndexDictionary["\(firstChar)"] {
+                        stops.append(busStop)
+                        sectionIndexDictionary["\(firstChar)"] = stops
                     } else {
-                        if var stops = sectionIndexDictionary["\(firstChar)"] {
-                            stops.append(busStop)
-                            sectionIndexDictionary["\(firstChar)"] = stops
-                        } else {
-                            sectionIndexDictionary["\(firstChar)"] = [busStop]
-                        }
+                        sectionIndexDictionary["\(firstChar)"] = [busStop]
                     }
                 }
             }
--- a/TCAT/Controllers/AllStopsTableViewController.swift
+++ b/TCAT/Controllers/AllStopsTableViewController.swift
@@ -64,45 +64,24 @@ class AllStopsTableViewController: UIViewController {
     }
 
     private func createSectionIndexesForBusStop() {
-        var currentChar: Character?
         var sectionIndexDictionary: [String: [Place]] = [:]
-        var currBusStopArray: [Place] = []
+        var numberBusStops: [Place] = []
 
-        currentChar = allStops.first?.name.capitalized.first
-
-        var numberBusStops: [Place] = {
-            if let firstStop = allStops.first {
-                return [firstStop]
-            }
-            return []
-        }()
-
-        if currentChar != nil {
+        if !allStops.isEmpty {
             for busStop in allStops {
-                if let firstChar = busStop.name.capitalized.first {
-                    if currentChar != firstChar {
-                        if CharacterSet.decimalDigits.contains(firstChar.unicodeScalars.first!) {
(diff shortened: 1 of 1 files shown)


TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -81,11 +81,14 @@ class AllStopsTableViewController: UIViewController {
             for busStop in allStops {
                 if let firstChar = busStop.name.capitalized.first {
                     if currentChar != firstChar {
+                        if CharacterSet.decimalDigits.contains(firstChar.unicodeScalars.first!) {
```
kevinchan159: can we safely unwrap instead of using `!`

kevinchan159 (approved):
LGTM w/ 1 minor comment!

TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -81,11 +81,14 @@ class AllStopsTableViewController: UIViewController {
             for busStop in allStops {
```
Omarrasheed: I know u didn't write this chunk of code but could we clean it up a little bit. `currentChar` seems to be unnecessary as we can use firstChar to filter into the correct letter. Let's remove the currentChar logic and just use if lets to check if the dictionary has an associated list with `firstChar`

Omarrasheed (changes requested):
Just a note about fixing up the code. I know you didn't write it but this would be a good chance to clean it up

TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -64,42 +64,24 @@ class AllStopsTableViewController: UIViewController {
...
-                if let firstChar = busStop.name.capitalized.first {
-                    if currentChar != firstChar {
-                        if !CharacterSet.decimalDigits.contains(currentChar!.unicodeScalars.first!) {
-                            sectionIndexDictionary["\(currentChar!)"] = currBusStopArray
-                            currBusStopArray = []
-                        }
-                        currentChar = firstChar
-                        currBusStopArray.append(busStop)
+                if let firstChar = busStop.name.capitalized.first, let firstScalar = firstChar.unicodeScalars.first {
```
Omarrasheed: yet another nit: can we put the second let on another line

TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -64,42 +64,24 @@ class AllStopsTableViewController: UIViewController {
...
-        var numberBusStops: [Place] = {
-            if let firstStop = allStops.first {
-                return [firstStop]
-            }
-            return []
-        }()
-
-        if currentChar != nil {
+        if !allStops.isEmpty {
             for busStop in allStops {
```
Omarrasheed: final nit i promise (for kevin): 
can we use 
```swift
allStops.forEach { busStop in 
 ...
}
```

TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -64,42 +64,24 @@ class AllStopsTableViewController: UIViewController {
...
+                    if CharacterSet.decimalDigits.contains(firstScalar) {
+                        numberBusStops.append(busStop)
                     } else {
-                        if CharacterSet.decimalDigits.contains(currentChar!.unicodeScalars.first!) {
-                            numberBusStops.append(busStop)
+                        if sectionIndexDictionary["\(firstChar)"] == nil {
+                            sectionIndexDictionary["\(firstChar)"] = [busStop]
                         } else {
-                            currBusStopArray.append(busStop)
+                            sectionIndexDictionary["\(firstChar)"]?.append(busStop)
```
Omarrasheed: nit to take out optional chaining: 
```swift
if let stops = sectionIndexDictionary["\(firstChar)"] {
  stops.append(busStop)
} else {
  sectionIndexDictionary["\(firstChar)"] = [busStop]
}
```

Omarrasheed (approved):
Beautiful :heart_eyes: :heart_eyes: :heart_eyes: :heart_eyes:

Just some v v minor comments

TCAT/Controllers/AllStopsTableViewController.swift
haiyingweng: if let makes stops immutable so should I do if var

Omarrasheed: Yea if var, my b

TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -64,42 +64,26 @@ class AllStopsTableViewController: UIViewController {
...
-                if let firstChar = busStop.name.capitalized.first {
-                    if currentChar != firstChar {
-                        if !CharacterSet.decimalDigits.contains(currentChar!.unicodeScalars.first!) {
-                            sectionIndexDictionary["\(currentChar!)"] = currBusStopArray
-                            currBusStopArray = []
-                        }
-                        currentChar = firstChar
-                        currBusStopArray.append(busStop)
+        if !allStops.isEmpty {
+            allStops.forEach { busStop in
```
young-k: Isn't the `if !allStops.isEmpty {` check meaningless here, since if `allStops` is empty, the `forEach` loop won't do anything?

young-k (commented):
In the future, make sure to resolve all comments that you fix in the "Conversation" tab!

TCAT/Controllers/AllStopsTableViewController.swift
haiyingweng: You right. Fixed!

Issue #316, opened 2019-10-05
Move live tracking logic out of route option cell into route options view controller


lucyxubroad:
Right now, each route options cell handles its own live tracking and cell updates, which is causing a lot of instabilities on the route options page such as flickering live statuses, live statuses on walking routes, and app crashing. Move the logic into the main view controller so that the view controller controls the rendering and live tracking of all cells. 

Connected issues: #286, #285

lucyxubroad:
Merged in #322

Issue #286, opened 2019-09-03
Flickering live tracking in routeOptionsCell


Omarrasheed:
Seems like the live tracking (specifically for the first route in the routeOptionsVC) tends to flicker on and off. Hopefully we can use this as an opportunity to refactor the way we handle live tracking updates for cells. We currently do a lot of weird stuff with times and the cells being the ones to manage their own timers, but if we move this to the VC handling the refetching of delays, we can potentially cut down on number of API calls and number of refreshes we make.

<img src="https://user-images.githubusercontent.com/36868927/64212730-4f19d080-ce78-11e9-9f13-5995ebc1e418.jpg" width="300">

lucyxubroad:
Resolved in #317

Issue #285, opened 2019-09-03
Live tracking on walking route


Omarrasheed:
### This should not be happening:
![Image from iOS (1)](https://user-images.githubusercontent.com/36868927/64212669-272a6d00-ce78-11e9-88b9-546ae438e4f9.jpg)

lucyxubroad:
Resolved in #317

PR #317, opened 2019-10-05
Lucy/redo live updates

lucyxubroad:
Addresses issues #316, #286, #285 
-  Moved out the live tracking logic from RouteTableViewCell into RouteOptionsViewController so that it can handle the retrieval of all live tracking for all cells and update the table together at once. 
- This solves our issue of app crashing, live labels showing up on walking routes, and live labels flickering.

Notes:
- Please pull down branch and double check that things do not break! I have done the scrolling around to make sure that app doesn't crash and I have also double checked that the routes I return are the same as the ones the production Transit returns.
- Please also take a look at the "JSONFileManager" stuff because I didn't really know what its doing / how to use it so I just adopted for my needs but I don't know how to test that its working as it should.
- Alanna has put up a new endpoint for all routes delays, which I can adapt to this one or we can try to push this one out first to address the crashing bugs?

### DIFF
--- a/TCAT/Cells/RouteTableViewCell.swift
+++ b/TCAT/Cells/RouteTableViewCell.swift
@@ -31,8 +31,6 @@ class RouteTableViewCell: UITableViewCell {
     private var routeDiagram: RouteDiagram!
     private let travelTimeLabel = UILabel()
     
-    var delay: DelayState?
-
     // MARK: Data vars
     private let containerViewLayoutInsets = UIEdgeInsets(top: 16, left: 16, bottom: 16, right: 12)
     private let networking: Networking = URLSession.shared.request
@@ -160,14 +158,18 @@ class RouteTableViewCell: UITableViewCell {
     }
 
     // MARK: Set Data
-    func configure(for route: Route, delegate: RouteTableViewCellDelegate? = nil, delayState: DelayState? = nil) {
+    func configure(for route: Route, delegate: RouteTableViewCellDelegate? = nil) {
         self.delegate = delegate
 
 //        timer = Timer.scheduledTimer(timeInterval: 5.0, target: self, selector: #selector(updateLiveElementsWithDelay(sender:)), userInfo: ["route": route], repeats: true)
         
-        if let delay = delayState {
-            setLiveElements(withDelayState: delay)
-        }
+//        if let delay = delayState {
+////            setLiveElements(withDelayState: delay)
+////            setDepartureTime(withStartTime: Date(), withDelayState: delay)
+//            setDepartureTimeAndLiveElements(withRoute: route, withDelay: delay)
+//        } else {
+//            setDepartureTimeAndLiveElements(withRoute: route, withDelay: DelayState.onTime(date: Date()))
+//        }
 
         setTravelTime(withDepartureTime: route.departureTime, withArrivalTime: route.arrivalTime)
 
@@ -193,15 +195,12 @@ class RouteTableViewCell: UITableViewCell {
 
     // MARK: Get Data
 
-    // Lucy - Getting if there is a delay
     private func getDelayState(fromRoute route: Route) -> DelayState {
-        
-        print("Getting delay state for \(route.endName)")
-        
+
         if let firstDepartDirection = route.getFirstDepartRawDirection() {
 
             let departTime = firstDepartDirection.startTime
-
+        
             if let delay = firstDepartDirection.delay {
                 let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
                 // Our live tracking only updates once every 30 seconds, so we want to show buses that are delayed by < 120 as on time in order to be more accurate about the status of slightly delayed buses. This way riders get to a bus stop earlier rather than later when trying to catch such buses.
@@ -229,12 +228,13 @@ class RouteTableViewCell: UITableViewCell {
             return
         }
 
-        //let delayState = getDelayState(fromRoute: route)
+        let delayState = getDelayState(fromRoute: route)
         
-        // Lucy - Update the view based on the delay state
-        //setDepartureTime(withStartTime: Date(), withDelayState: delayState)
-        //setLiveElements(withDelayState: delayState)
+        setDepartureTime(withStartTime: Date(), withDelayState: delayState)
+        setLiveElements(withDelayState: delayState)
     }
+    
+    // Note for Lucy: "route.getFirstDepartRawDirection()?.delay = delay" missing!
 
 //    @objc private func updateLiveElementsWithDelay(sender: Timer) {
 //        guard let userInfo = sender.userInfo as? [String: Route],
@@ -300,31 +300,25 @@ class RouteTableViewCell: UITableViewCell {
     }
 
     // Update all of the live elements based on the delay state.
-    private func setLiveElements(withDelayState: DelayState) {
+    private func setLiveElements(withDelayState delayState: DelayState) {
         
-        if let delayState = delay {
-            
-            print("Yes delay")
-            print(delayState)
-            
-            switch delayState {
-            case .late(date: let delayedDepartureTime):
-                liveLabel.textColor = Colors.lateRed
-                liveLabel.text = "Late - \(Time.timeString(from: delayedDepartureTime))"
-                liveIndicatorView.setColor(to: Colors.lateRed)
-                showLiveElements()
+        switch delayState {
+        case .late(date: let delayedDepartureTime):
+            liveLabel.textColor = Colors.lateRed
+            liveLabel.text = "Late - \(Time.timeString(from: delayedDepartureTime))"
+            liveIndicatorView.setColor(to: Colors.lateRed)
+            showLiveElements()
 
-            case .onTime(date: _):
-                liveLabel.textColor = Colors.liveGreen
-                liveLabel.text = "On Time"
-                liveIndicatorView.setColor(to: Colors.liveGreen)
-                showLiveElements()
+        case .onTime(date: _):
+            liveLabel.textColor = Colors.liveGreen
+            liveLabel.text = "On Time"
+            liveIndicatorView.setColor(to: Colors.liveGreen)
+            showLiveElements()
 
-            case .noDelay(date: _):
-                hideLiveElements()
-            }
+        case .noDelay(date: _):
+            hideLiveElements()
         }
-
+        
     }
 
     private func showLiveElements() {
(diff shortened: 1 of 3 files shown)


TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,65 @@ class RouteOptionsViewController: UIViewController {
...
+                    let stopId = direction.stops.first?.id {
+                    getDelay(tripId: tripId, stopId: stopId).observe(with: { result in
+                        let fileName = "RouteTableViewCell"
+                        DispatchQueue.main.async {
+                            switch result {
+                            case .value (let delayResponse):
+                                guard (delayResponse.data != nil), let delay = delayResponse.data else {
+                                    return
+                                }
+                                // LUCY - Check
```
young-k: should this be a TODO?

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,65 @@ class RouteOptionsViewController: UIViewController {
...
+        routeResults.reloadData()
+    }
+    
+    private func getDelay(tripId: String, stopId: String) -> Future<Response<Int?>> {
+        return networking(Endpoint.getDelay(tripID: tripId, stopID: stopId)).decode()
+    }
+
+    @objc func updateLiveTracking(sender: Timer) {
+        for routesArray in routes {
+            for (index, route) in routesArray.enumerated() {
```
young-k, line 339: just a little confused why we have two nested loops here –– is `routesArray` corresponding to a single route, and `(index, route)` covers the case of transfers? this might be useful to have in a comment, just to make it clearer for anyone in the future!

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,65 @@ class RouteOptionsViewController: UIViewController {
...
+                                }
+                                let departTime = direction.startTime
+                                let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
+                                let isLateDelay = (Time.compare(date1: delayedDepartTime, date2: departTime) == .orderedDescending)
+                                if isLateDelay {
+                                    let delayState = DelayState.late(date: delayedDepartTime)
+                                    self.delayDictionary[route.routeId] = delayState
+                                } else {
+                                    let delayState = DelayState.onTime(date: departTime)
+                                    self.delayDictionary[route.routeId] = delayState
```
young-k: can't this line be moved out of the `if / else` since it occurs in both cases?

this logic also seems like it could be simplified, do we need to keep track of both `departTime` and the `delayedDepartTime`, or do we only need to keep track of one?

another way to clean this up might be to have `DelayState` take a `time` and an `enum` representing the state, so that we don't have to call two separate constructors for the same use case!

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,65 @@ class RouteOptionsViewController: UIViewController {
...
+                                }
+                                route.getFirstDepartRawDirection()?.delay = delay // LUCY - Check
```
young-k: same question regarding a possible `TODO` here?

young-k (commented):
Not sure that I know better than @kevinchan159 or @Omarrasheed, but I think having two separate PRs for this would be better, just to keep PR size down!

TCAT/Cells/RouteTableViewCell.swift
```diff
@@ -189,6 +182,7 @@ class RouteTableViewCell: UITableViewCell {
     // MARK: Get Data
 
     private func getDelayState(fromRoute route: Route) -> DelayState {
+
```
kevinchan159: nit: delete newline

TCAT/Cells/RouteTableViewCell.swift
```diff
@@ -213,6 +207,7 @@ class RouteTableViewCell: UITableViewCell {
     }
 
     private func setDepartureTimeAndLiveElements(withRoute route: Route) {
+
```
kevinchan159: same

TCAT/Controllers/RouteOptionsViewController+Extensions.swift
```diff
@@ -208,9 +208,11 @@ extension RouteOptionsViewController: UITableViewDataSource {
 
         guard let cell = tableView.dequeueReusableCell(withIdentifier: Constants.Cells.routeOptionsCellIdentifier, for: indexPath) as? RouteTableViewCell
             else { return UITableViewCell() }
-
-        cell.configure(for: routes[indexPath.section][indexPath.row], delegate: self)
-
+        
+        let route_id = routes[indexPath.section][indexPath.row].routeId
```
kevinchan159: `routeId`

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,65 @@ class RouteOptionsViewController: UIViewController {
...
+                if !route.isRawWalkingRoute(),
+                    let direction = route.getFirstDepartRawDirection(),
+                    let tripId = direction.tripIdentifiers?.first,
+                    let stopId = direction.stops.first?.id {
+                    getDelay(tripId: tripId, stopId: stopId).observe(with: { result in
+                        let fileName = "RouteTableViewCell"
+                        DispatchQueue.main.async {
+                            switch result {
+                            case .value (let delayResponse):
+                                guard (delayResponse.data != nil), let delay = delayResponse.data else {
```
kevinchan159: nit: remove `( )`

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,65 @@ class RouteOptionsViewController: UIViewController {
...
+                                    return
+                                }
+                                // LUCY - Check
+                                let isNewDelayValue = (route.getFirstDepartRawDirection()?.delay != delay)
```
kevinchan159: remove parens

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,65 @@ class RouteOptionsViewController: UIViewController {
...
+        return networking(Endpoint.getDelay(tripID: tripId, stopID: stopId)).decode()
+    }
+
+    @objc func updateLiveTracking(sender: Timer) {
+        for routesArray in routes {
+            for (index, route) in routesArray.enumerated() {
+                if !route.isRawWalkingRoute(),
+                    let direction = route.getFirstDepartRawDirection(),
+                    let tripId = direction.tripIdentifiers?.first,
+                    let stopId = direction.stops.first?.id {
```
kevinchan159: in order to prevent the indentation from getting too nested, can we switch to using a `guard let` here instead?

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,65 @@ class RouteOptionsViewController: UIViewController {
...
+                                    JSONFileManager.shared.logURL(timestamp: Date(), urlName: "Delay requestUrl", url: Endpoint.getDelayUrl(tripId: tripId, stopId: stopId))
+                                    if let data = try? JSONEncoder().encode(delayResponse) {
+                                        do { try JSONFileManager.shared.saveJSON(JSON.init(data: data), type: .delayJSON(rowNum: index)) } catch let error {
+                                            let line = "\(fileName) \(#function): \(error.localizedDescription)"
+                                            print(line)
+                                        }
+                                    }
+                                }
+                                let departTime = direction.startTime
+                                let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
```
kevinchan159: might be wrong but from how i'm understanding it, as long as `delay > 0`, then we have a delay right? if so, we can simplify the logic a bit to:
```
if delay > 0 {
  let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
  let delayState = DelayState.late(date: delayedDepartTime)
} else {
  let delayState = let delayState = DelayState.onTime(date: departTime)
}
```

kevinchan159 (commented):
🥳
as for incorporating Alanna's new endpoint, I would say to do that in a separate PR after we merge this in

TCAT/Controllers/RouteOptionsViewController.swift
lucyxubroad: Looking at the code, it seems that this should be true but I'm doing some print statements and it seems like for small delay numbers (<35), we still consider that as on time even though its > 0? I believe this may have to do with the fact that our delay checks happen on a timer so some values may be a bit off. I think Omar would have to confirm though. @Omarrasheed

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -320,7 +321,68 @@ class RouteOptionsViewController: UIViewController {
...
+                                    do { try JSONFileManager.shared.saveJSON(JSON.init(data: data), type: .delayJSON(rowNum: index)) } catch let error {
+                                        let line = "\(fileName) \(#function): \(error.localizedDescription)"
+                                        print(line)
+                                    }
+                                }
+                            }
+                            let departTime = direction.startTime
+                            let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
+                            var delayState: DelayState!
+                            let isLateDelay = (Time.compare(date1: delayedDepartTime, date2: departTime) == .orderedDescending)
```
young-k: nit, but can't we delete the outermost `()` here?

TCAT/Controllers/RouteOptionsViewController+Extensions.swift
```diff
@@ -208,9 +208,11 @@ extension RouteOptionsViewController: UITableViewDataSource {
 
         guard let cell = tableView.dequeueReusableCell(withIdentifier: Constants.Cells.routeOptionsCellIdentifier, for: indexPath) as? RouteTableViewCell
             else { return UITableViewCell() }
-
-        cell.configure(for: routes[indexPath.section][indexPath.row], delegate: self)
-
+        
+        let routeId = routes[indexPath.section][indexPath.row].routeId
+        
+        cell.configure(for: routes[indexPath.section][indexPath.row], delayState: delayDictionary[routeId])
```
young-k: nit, but can't we save `routes[indexPath.section][indexPath.row]` to a variable, rather than the `routeId`, and that'll be cleaner... so:

```
let a = routes[indexPath.section][indexPath.row]
cell.configure(for: a, delayState: delayDictionary[a.routeId])
```

young-k (approved):
lgtm

TCAT/Controllers/RouteOptionsViewController.swift
kevinchan159: ah yeah i think we do that because our "live tracking" isn't actually live tracking since TCAT only gives us their updated bus locations every 30 seconds

kevinchan159 (approved):
🚀 🎢 💯

PR #322, opened 2019-10-11
Lucy/new delays endpoint

lucyxubroad:
**Addresses Issue #316 and Is an Extension of my Previous PR #317** 

_Background:_
- Issue was to move all live tracking logic out of the route cells into the route options view controller because having the logic in the cell was causing a lot of instabilities and crashes within our app
- This PR extends #317, where I initially moved the live tracking logic into the routes option view controller using our previous /delay endpoint to retrieve delays for each route
- For this PR, I updated my code to now use Alanna's new /delays endpoint so that I get delays for all routes at once

_New Things:_
- I added a new /delays route to our networking files and also created a trip dictionary so that we can map from route to trip and back. 
- I updated the way we did JSON logging so that we use routeId as opposed to rowNum because rowNum is pretty arbitrary to how our table renders the routes.

_Some Questions:_
- Right now, I'm rendering my trips array inside of my "updateAllRoutesLiveTracking" function, which fires every 5 seconds. Should I move that into our searchForRoutes() function instead so that the trips array is instantiated the same time our routes are?
- I am currently updating table every 20 seconds and fetching delays every 5 seconds. Let me know if I should change those numbers!

### DIFF
--- a/TCAT/Controllers/RouteOptionsViewController.swift
+++ b/TCAT/Controllers/RouteOptionsViewController.swift
@@ -375,65 +375,11 @@ class RouteOptionsViewController: UIViewController {
                             route.getFirstDepartRawDirection()?.delay = delay
                         }
                     case .error(let error):
-                        print(error)
+                        self.printClass(context: "\(#function) error", message: error.localizedDescription)
                     }
                 }
             })
     }
-
-//    @objc func updateLiveTracking(sender: Timer) {
-        // For each route in each route array inside of the 'routes' array,
-        // retrieve its delay. Use index of route to save delay for route to
-        // JSON file.
-//        updateAllRoutesLiveTracking()
-//        for routesArray in routes {
-//            for (index, route) in routesArray.enumerated() {
-//                if !route.isRawWalkingRoute() {
-//                    guard let direction = route.getFirstDepartRawDirection(),
-//                        let tripId = direction.tripIdentifiers?.first,
-//                        let stopId = direction.stops.first?.id else {
-//                            return
-//                    }
-//                    getDelay(tripId: tripId, stopId: stopId).observe(with: { result in
-//                        let fileName = "RouteTableViewCell"
-//                        DispatchQueue.main.async {
-//                            switch result {
-//                            case .value (let delayResponse):
-//                                guard delayResponse.data != nil, let delay = delayResponse.data else {
-//                                    return
-//                                }
-//                                let isNewDelayValue = route.getFirstDepartRawDirection()?.delay != delay
-//                                if isNewDelayValue {
-//                                    JSONFileManager.shared.logDelayParemeters(timestamp: Date(), stopId: stopId, tripId: tripId)
-//                                    JSONFileManager.shared.logURL(timestamp: Date(), urlName: "Delay requestUrl", url: Endpoint.getDelayUrl(tripId: tripId, stopId: stopId))
-//                                    if let data = try? JSONEncoder().encode(delayResponse) {
-//                                        do { try JSONFileManager.shared.saveJSON(JSON.init(data: data), type: .delayJSON(rowNum: index)) } catch let error {
-//                                            let line = "\(fileName) \(#function): \(error.localizedDescription)"
-//                                            print(line)
-//                                        }
-//                                    }
-//                                }
-//                                let departTime = direction.startTime
-//                                let delayedDepartTime = departTime.addingTimeInterval(TimeInterval(delay))
-//                                var delayState: DelayState!
-//                                let isLateDelay = Time.compare(date1: delayedDepartTime, date2: departTime) == .orderedDescending
-//                                if isLateDelay {
-//                                    delayState = DelayState.late(date: delayedDepartTime)
-//                                } else {
-//                                    delayState = DelayState.onTime(date: departTime)
-//                                }
-//                                self.delayDictionary[route.routeId] = delayState
-//                                route.getFirstDepartRawDirection()?.delay = delay
-//
-//                            case .error (let error):
-//                                print(error)
-//                            }
-//                        }
-//                    })
-//                }
-//            }
-//        }
-//    }
     
     @objc private func refreshRoutesAndTime() {
         let now = Date()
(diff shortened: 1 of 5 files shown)


TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -113,14 +117,14 @@ class RouteOptionsViewController: UIViewController {
...
-        routeTimer = Timer.scheduledTimer(timeInterval: 5.0, target: self, selector: #selector(updateLiveTracking(sender:)), userInfo: nil, repeats: true)
+        routeTimer = Timer.scheduledTimer(timeInterval: 5.0, target: self, selector: #selector(updateAllRoutesLiveTracking(sender:)), userInfo: nil, repeats: true)
         updateTimer = Timer.scheduledTimer(timeInterval: 20.0, target: self, selector: #selector(rerenderLiveTracking(sender:)), userInfo: nil, repeats: true)
 
         // Check for 3D Touch availability
         if traitCollection.forceTouchCapability == .available {
             registerForPreviewing(with: self, sourceView: view)
         }
-
+
```
kevinchan159: nit: delete whitespace

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -327,36 +331,50 @@ class RouteOptionsViewController: UIViewController {
...
                 }
-                getDelay(tripId: tripId, stopId: stopId).observe(with: { result in
-                    let fileName = "RouteTableViewCell"
-                    DispatchQueue.main.async {
-                        switch result {
-                        case .value (let delayResponse):
-                            guard delayResponse.data != nil, let delay = delayResponse.data else {
-                                return
+            }
+        }
```
kevinchan159: why do we have to recreate this `trips` array everytime this timer is triggered rather than just doing it once?

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -49,6 +49,7 @@ class RouteOptionsViewController: UIViewController {
     var searchTo: Place!
     var searchType: SearchBarType = .to
     var showRouteSearchingLoader: Bool = false
+    var trips: [Trip] = []
```
danielVebman, line 52: What is this variable used for? I can't find any instances where it is written to or read in `RouteOptionsViewController`. In `updateAllRoutesLiveTracking(sender:)` you seem to recreate `trips` only locally.

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -327,36 +331,50 @@ class RouteOptionsViewController: UIViewController {
...
+        return networking(Endpoint.getAllDelays(trips: trips)).decode()
     }
-
-    @objc func updateLiveTracking(sender: Timer) {
-        // For each route in each route array inside of the 'routes' array,
-        // retrieve its delay. Use index of route to save delay for route to
-        // JSON file.
+    
+    @objc func updateAllRoutesLiveTracking(sender: Timer) {
+        var trips: [Trip] = []
```
danielVebman: Is this supposed to be related to the `trip` class variable?

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -327,36 +331,50 @@ class RouteOptionsViewController: UIViewController {
...
-                guard let direction = route.getFirstDepartRawDirection(),
-                    let tripId = direction.tripIdentifiers?.first,
-                    let stopId = direction.stops.first?.id else {
-                        return
+            for route in routesArray {
+                if !route.isRawWalkingRoute() {
+                    guard let direction = route.getFirstDepartRawDirection(),
+                        let tripId = direction.tripIdentifiers?.first,
+                        let stopId = direction.stops.first?.id else {
+                            return
```
danielVebman: Are you sure you want to `return` here and not `continue`? Return will completely stop the update. Continue will just stop processing that one route and keep trying with the next ones.

Also, a note about the if statement (`if !route.isRawWalkingRoute()`): That method already has a ton of nested blocks. You can instead replace the if statement with the following and remove some indentation:

    guard !route.isRawWalkingRoute else { continue }

TCAT/Controllers/RouteOptionsViewController.swift
lucyxubroad: oh ya, this was also a concern from when I was writing this too. I think it was mainly cause I wanted to make sure that we always had the updated routes array to render our trips array from and I wasn't sure when routes were generated but I moved it out so this should be fixed now!

TCAT/Controllers/RouteOptionsViewController.swift
lucyxubroad: oops, this was from when I was considering rendering trips globally and not locally! I moved out where I render all the trips so now im using this variable as opposed to my local one.

TCAT/Controllers/RouteOptionsViewController.swift
lucyxubroad: I moved out the for loop so now I use the global trips variable as opposed to the local one!

TCAT/Controllers/RouteOptionsViewController.swift
lucyxubroad: good catch! thanks Daniel!

danielVebman (approved):

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -503,6 +501,25 @@ class RouteOptionsViewController: UIViewController {
             }
         }
     }
+    
+    private func getRoutesTrips() {
+        // For each route in each route array inside of the 'routes' array, get its
+        // tripId and stopId to create trip array for request to get all delays.
+        for routesArray in routes {
+            for route in routesArray {
+                guard (!route.isRawWalkingRoute()) else { continue }
```
kevinchan159: remove parens around condition

kevinchan159: can't we also combine this guard with the guard statement on the next line?

kevinchan159 (approved):
no more crashing!!! 🥳

Issue #305, opened 2019-09-24
Standardize logging


kevinchan159:
Right now, most logs look like:
```
print("[FavoritesTableViewController] blah blah blah")
```
Ideally there should be some helper method that follows this format of `print([Class Name] message)` without us having to always type the class name

PR #323, opened 2019-10-12
Standardize logging with print statement

haiyingweng:
For #305 

Standardize logging with print statement using a helper method. We don't need to always put in class names. 

Prints something like this: 
Screen Shot 2019-10-11 at 8 02 09 PM

### DIFF
--- a/TCAT/Controllers/AllStopsTableViewController.swift
+++ b/TCAT/Controllers/AllStopsTableViewController.swift
@@ -41,6 +41,7 @@ class AllStopsTableViewController: UIViewController {
         setupConstraints()
 
         refreshAllStops()
+        Analytics.shared.logWithPrintStatement(currentClass: self, context: "\(#function) error", message: "some error")
     }
 
     private func setupTableView() {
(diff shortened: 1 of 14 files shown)


Omarrasheed (approved):
Love it 😍

danielVebman (approved):
This is cool. Would this be valuable as an NSObject extension? This would be it:

```swift
extension NSObject {
    func logWithPrintStatement(context: String, message: String) {
        Analytics.shared.logWithPrintStatement(currentClass: Self.self, context: context, message: message)
    }
}
```

That way you don't have to say `Analytics.shared. ...` every time.

Not super important, but `logWithPrintStatement` is pretty wordy. Does the method actually *log* anything or does it just print to the console? Why not call it `printClass(_:context:message:)` or just `print(...)`?

haiyingweng:
@danielVebman 
Yea, I think it is actually a good idea to put the helper method in  NSObject extension. In that case, I don't need to put it in Analytics anymore since it doesn't actually log anything.

Issue #267, opened 2019-08-26
Excess calls in didUpdateLocation in RouteDetail+ContentVC


Omarrasheed:
Seems like we're running our private function `didUpdateLocation()` way more than needed. Let's try and find a way to cut down on these calls

PR #324, opened 2019-10-14
Daniel/did update location excess calls

danielVebman:
This is a solution to issue #267. I made `isInitialUpdate` private(set) instead of private so it is accessible from extensions. The two location manager delegate methods that call `didUpdateLocation` now check if `isInitialUpdate` first. The if statements should reduce overhead compared to function calls, but honestly I'm not sure if it is a meaningful difference from a performance perspective.

Also, I'm not sure why the commits on my last branch are in this pull request. I thought they were already squashed and I'm on a new branch.

### DIFF
--- a/TCAT/Controllers/FavoritesTableViewController.swift
+++ b/TCAT/Controllers/FavoritesTableViewController.swift
@@ -10,6 +10,11 @@ import UIKit
 import DZNEmptyDataSet
 import FutureNova
 
+protocol FavoritesSelectionDelegate: class {
+    /** Indicates to `HomeMapViewController` that it should reload its table. */
+    func didAddNewFavorite()
+}
+
 class FavoritesTableViewController: UIViewController {
 
     private var searchBar = UISearchBar()
@@ -174,9 +179,4 @@ extension FavoritesTableViewController: UISearchBarDelegate {
         }
     }
 
-}
-
-protocol FavoritesSelectionDelegate: class {
-    /** Indicates to `HomeMapViewController` that it should reload its table. */
-    func didAddNewFavorite()
 }
--- a/TCAT/Controllers/FavoritesTableViewController.swift
+++ b/TCAT/Controllers/FavoritesTableViewController.swift
@@ -22,7 +22,9 @@ class FavoritesTableViewController: UIViewController {
             tableView.reloadData()
         }
     }
-
+    
+    weak var selectionDelegate: FavoritesSelectionDelegate?
+    
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -118,6 +120,7 @@ extension FavoritesTableViewController: UITableViewDelegate {
         tableView.deselectRow(at: indexPath, animated: true)
         if let place = resultsSection.getItem(at: indexPath.row) {
             Global.shared.insertPlace(for: Constants.UserDefaults.favorites, place: place, bottom: true)
+            selectionDelegate?.didAddNewFavorite()
             dismissVC()
         }
     }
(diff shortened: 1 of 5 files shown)


TCAT/Controllers/HomeOptionsCardViewController+Extensions.swift
```diff
@@ -280,3 +280,9 @@ extension HomeOptionsCardViewController: UITableViewDelegate {
         }
     }
 }
+
+extension HomeOptionsCardViewController: FavoritesSelectionDelegate {
+    func didAddNewFavorite() {
+        updatePlaces()
+    }
+}
```
kevinchan159: can we add a newline right after line 284 and right before 288

kevinchan159 (approved):
looks great! remember to also request reviews from @lucyxubroad, @haiyingweng, as well as your external reviewers!
regarding your old commits appearing, try doing:
```
git checkout master (switch branches to master)
git pull (pull changes from the master branch on github to your local master branch)
git checkout [your branch name] (switch back to your branch)
git rebase master (apply changes)
```
and then commit/push again.

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -459,8 +459,8 @@ class RouteDetailContentViewController: UIViewController {
     /// Completion after locationManager functions return
     func didUpdateLocation() {
         // TODO #267: Find better way to cut down on didUpdateLocation calls
```
lucyxubroad: Take this out before you merge this PR!

lucyxubroad (approved):
looks pretty good to me!

TCAT/Controllers/FavoritesTableViewController.swift
```diff
@@ -10,6 +10,11 @@ import UIKit
 import DZNEmptyDataSet
 import FutureNova
 
+protocol FavoritesSelectionDelegate: class {
+    /** Indicates to `HomeMapViewController` that it should reload its table. */
```
young-k: hm, should we be commenting like this for single line comments? @kevinchan159

young-k:
This PR title should also optimally be changed prior to merging!

TCAT/Controllers/FavoritesTableViewController.swift
lucyxubroad: We seem to have pretty inconsistent comment styles throughout Transit in general btw. Should probably standardize going forward.

danielVebman: For a single line doc comment that should be `///` actually. I'll change that.

young-k: Shouldn't this be `//`? `///` are typically used for docstrings, no?

young-k (changes requested):
Few changes –– the branch should be renamed, and commits should be in present tense (it doesn't matter since they'll be squashed, but it's important to follow the standard). You can rename the branch with `git branch -m`, and commits can be edited by rebasing!

TCAT/Controllers/FavoritesTableViewController.swift
danielVebman: @young-k I had intended that as a docstring so that others on the team could figure out what that function's purpose is without navigating to the file. That might make more sense as a comment though. I'll change it if you think a comment is more appropriate.

Edit: It actually might make make most sense to just remove the comment altogether. I'm not sure how explicit one is supposed to be with comments.

danielVebman:
@young-k what should the branch be named? Is `daniel/reduce-calls-to-didUpdateLocation` good?

PR #326, opened 2019-10-17
Update to 6.1.4

Omarrasheed:
## Overview

Updated build numbers + small maintenance

## Changes Made

Just updated the build numbers, update pod (deleted podfile.lock and ran pod install), and ran swiftlint

## Test Coverage

Doesn't break + shipped to testflight


kevinchan159 (approved):
sheep it

Issue #306, opened 2019-09-24
Standardize handling of errors


kevinchan159:
We should have some formal convention of what to do and log when there is an error.

Issue #171, opened 2018-10-15
Add and improve Analytics, especially for errors


mattbarker016:
Add additional analytics for any location in the app that we print error messages. We don't know if these are firing in the wild or not, and this will help us gain better understanding of the app's usage. This should be standardized to a generic ErrorPayload that falls under the same Fabric Event.

PR #327, opened 2019-10-17
Standardize error handling

haiyingweng:
## Overview
For most errors, we weren't really doing anything to them except printing them. This PR adds a payload to log errors to Firebase so we can have a better understanding of where errors occur.  

## Changes Made
Added a new payload called GetErrorPayload to log errors relating to getSomething.

**Question** 
Not sure if GetErrorPayload is a good name. I didn't go with NetworkErrorPayload since not all errors are actually network error. Lmk if there could be a better name

## Test Coverage
The app still runs 

## Related PRs or Issues
#306 
#171

### DIFF
--- a/TCAT/Controllers/AllStopsTableViewController.swift
+++ b/TCAT/Controllers/AllStopsTableViewController.swift
@@ -152,7 +152,7 @@ class AllStopsTableViewController: UIViewController {
                             userDefaults.set(encodedObject, forKey: Constants.UserDefaults.allBusStops)
                         } catch let error {
                             self.printClass(context: "\(#function) error", message: error.localizedDescription)
-                            let payload = GetErrorPayload(
+                            let payload = NetworkErrorPayload(
                                 location: "\(self) Get All Stops",
                                 type: "\((error as NSError).domain)",
                                 description: error.localizedDescription)
@@ -164,7 +164,7 @@ class AllStopsTableViewController: UIViewController {
                     }
                 case .error(let error):
                     self.printClass(context: "\(#function) error", message: error.localizedDescription)
-                    let payload = GetErrorPayload(
+                    let payload = NetworkErrorPayload(
                         location: "\(self) Get All Stops",
                         type: "\((error as NSError).domain)",
                         description: error.localizedDescription)
--- a/TCAT/Controllers/AllStopsTableViewController.swift
+++ b/TCAT/Controllers/AllStopsTableViewController.swift
@@ -41,7 +41,6 @@ class AllStopsTableViewController: UIViewController {
         setupConstraints()
 
         refreshAllStops()
-        print("\(self)")
     }
 
     private func setupTableView() {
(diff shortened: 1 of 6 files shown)


TCAT/Controllers/AllStopsTableViewController.swift
```diff
@@ -152,13 +152,17 @@ class AllStopsTableViewController: UIViewController {
                             userDefaults.set(encodedObject, forKey: Constants.UserDefaults.allBusStops)
                         } catch let error {
                             self.printClass(context: "\(#function) error", message: error.localizedDescription)
+                            let payload = GetErrorPayload(location: "\(self) Get All Stops", type: "\((error as NSError).domain)", description: error.localizedDescription, url: nil)
```
kevinchan159: can we separate these across multiple lines like so:
```
let payload = GetErrorPayload(
  location: ,
  type: ,
  ...
)
```

kevinchan159 (approved):
nice work!

TCAT/Controllers/AllStopsTableViewController.swift
Omarrasheed: Let's do this for all

TCAT/Utilities/Analytics.swift
```diff
@@ -218,3 +208,14 @@ struct WhatsNewCardDismissedPayload: Payload {
...
+
+/// Log get errors
+struct GetErrorPayload: Payload {
+    static let eventName: String = "Get Error"
+    let deviceInfo = DeviceInfo()
+    
+    let location: String
+    let type: String 
+    let description: String
+    let url: String?
```
Omarrasheed: Let's take our `url` as well. We currently store `requestURL` to be able to write it to log files that only get used if the user submits feedback to the app. Unfortunately this is all like v legacy code and tbh we like never look at it. Idt the issues our users end up running into r complex enough where we can't repro and need specific logs for it. I'll make an issue to scrap all of the JSONFileManager stuff that uses the url but for now, we can just omit this field on the payload object

Omarrasheed (changes requested):
lmfao love the test plan

TCAT/Utilities/Analytics.swift
```diff
@@ -218,3 +208,13 @@ struct WhatsNewCardDismissedPayload: Payload {
 
     let actionDescription: String
 }
+
+/// Log get errors
+struct GetErrorPayload: Payload {
+    static let eventName: String = "Get Error"
```
Omarrasheed: Can we change the name of this to be `NetworkError` and the eventName to `Network Error`

TCAT/Utilities/Analytics.swift
```diff
@@ -218,3 +208,13 @@ struct WhatsNewCardDismissedPayload: Payload {
...
+    let deviceInfo = DeviceInfo()
+
```
Omarrasheed, line 216: Final nit i promise but we can remove this newline

Omarrasheed (approved):
Last few small comments but looks great!

TCAT/Utilities/Analytics.swift
haiyingweng: all the other payloads had a line after `deviceInfo` so i thought i would keep it consistent, or should i remove all the lines for the other payloads?

TCAT/Utilities/Analytics.swift
haiyingweng: I didn't go with `NetworkErrorPayload` bc not all of the errors were actually network error. For example, in `RouteOptionsVC` there's an error in getting routes for when place coordinates are not valid.

TCAT/Utilities/Analytics.swift
Omarrasheed: Hmm we can j keep then no problem

TCAT/Utilities/Analytics.swift
Omarrasheed: Hmm i see. Lets still change it to `NetworkErrorPayload` because even tho some of the errors aren't specifically using the result of a network request, they still have to do with the information being passed around during network requests.

Omarrasheed (approved):

PR #333, opened 2019-10-23
Standardize comments and other formatting issues

kevinchan159:
## Overview
Fix formatting issues with comments, indentation, spaces, and other minor issues.

## Changes Made

- Fixed comments to distinguish between `///` and `//`. `//` should only be used in function comments.
- Removed all instances of `/** /`
- Fixed indentation for lines that are too long
- Add/remove newlines
- Fixed `// MARK` comments

## Test Coverage

Ran the app

### DIFF
--- a/TCAT/AppDelegate.swift
+++ b/TCAT/AppDelegate.swift
@@ -17,7 +17,7 @@ import SwiftyJSON
 import UIKit
 import WhatsNewKit
 
-// This is used for app-specific preferences
+/// This is used for app-specific preferences
 let userDefaults = UserDefaults.standard
 
 @UIApplicationMain
@@ -151,7 +151,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Called when the application is about to terminate. Save data if appropriate. See also applicationDidEnterBackground:.
     }
     
-    // MARK: Helper Functions
+    // MARK: - Helper Functions
     
     /// Convert BusStop and PlaceResult models to new unified Place model.
     func migrationToNewPlacesModel(completion: @escaping (_ success: Bool, _ errorDescription: String?) -> Void) {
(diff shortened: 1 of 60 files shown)


haiyingweng (approved):
Wow! :clap:

TCAT/AppDelegate.swift
```diff
@@ -151,7 +151,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         // Called when the application is about to terminate. Save data if appropriate. See also applicationDidEnterBackground:.
     }
     
-    // MARK: Helper Functions
+    // MARK: - Helper Functions
```
haiyingweng, line 154: Question: 
When is `MARK:` used usually?

kevinchan159: its usually used to denote a section in a file. when you add `// MARK: -` in a file, XCode will allow you to directly jump to that section quickly in the IDE.

TCAT/Views/SummaryView.swift
```diff
@@ -27,13 +27,11 @@ class SummaryView: UIView {
...
-         This value ends up getting overwritten by constraints, which is what we want,
-         but for some reason if it is not set prior to writing the constraints, the
-         entire view comes out blank. I'm still investigating but it seems to be an,
-         issue with the Pulley Pod that we're using.
-        */
+        // TODO:
+        // This value ends up getting overwritten by constraints, which is what we want,
+        // but for some reason if it is not set prior to writing the constraints, the
+        // entire view comes out blank. I'm still investigating but it seems to be an,
+        // issue with the Pulley Pod that we're using.
```
lucyxubroad, line 34: What is the reason that we don't want to use /**/ for multiline comments?

kevinchan159: i'm following Google's style guidelines and `/** */` is used for Java docs so we shouldn't use them for Swift

PR #334, opened 2019-10-23
Pop up search view controller in searchTo and searchFrom

lucyxubroad:
## Overview

With iOS 13 changes, our search view controller is no longer popping up when a user taps on search to / search from. This PR tries to fix that problem. 

## Changes Made

It seemed that our search controller is not being set active where we want it to. I did research and this seems to be an issue that other people are also having. A fix for now is to wrap an async after around the code so that we set the search controller to be active right after the search bar shows up. 

## Test Coverage

I pressed on the bar items to make sure the table view pops up. 

## Next Steps

We may have to potentially redo our search bar logic overall? There seems to be some layout issues with our hiding and showing of the bar right now.. But that's going to be quite a big endeavor I think. 

  

</details>


TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -317,13 +317,15 @@ class RouteOptionsViewController: UIViewController {
...
         // After adding searchController to navigation bar, we need to call
-        // setNeedsLayout in order to for the navigation bar to increase its height
+        // setNeedsLayout in order for the navigation bar to increase its height
         // to account for the search bar
         navigationController?.view.setNeedsLayout()
         backButton = navigationItem.leftBarButtonItem
         navigationItem.setLeftBarButton(nil, animated: false)
         navigationItem.hidesBackButton = true
-        searchBarView.searchController?.isActive = true
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.03) {
```
haiyingweng: what does the deadline thing do?

lucyxubroad: Without it, our searchController does not get set to active. Adding the deadline will asynchronously set the searchController to isActive true after a set time.

haiyingweng (approved):
looks good based on the demo you showed me

danielVebman (approved):
Looks good. Maybe later we can figure out a way to do this without async dispatch.

kevinchan159 (approved):
i'm not the biggest fan of something like this but it seems like the reason we need something like this is because the searching logic is kind of jank as a whole. could we try to tackle this in the refactor @danielVebman

Omarrasheed (approved):
Seems jank but lets j push it lmao

Issue #335, opened 2019-10-24
Log index of selected search result


Omarrasheed:
We want to start logging the searches that users make and the corresponding index of the search result they choose. The reason behind adding this is to see if we have reason to implement some very basic _ranking_ of search results on backend to provide users with more relevant results.

PR #336, opened 2019-10-27
Log selected search

haiyingweng:
## Overview

Log user's search and the index of the selected result 

## Changes Made

Created `SearchResultSelectedPayload` and added payload wherever user selects search result 

## Test Coverage

Used print statements to make sure `searchText` and `selectedIndex` are correct

## Related PRs or Issues 

#335


danielVebman (approved):
Looks good to me. It is somewhat repetitive, so maybe you can refactor some parts, but that might not be meaningfully better than what you have right now.

lucyxubroad (approved):
I agree with what Daniel said.

Omarrasheed:
Can we also add total number of results? Cuz if the user picks 1/1 it’s much different from 1/10. Will also help us gauge if people are feeling the need to type out the whole thing or not. Not sure it’ll have like a lot of significance but it won’t hurt to also track. Otherwise LGTM

Issue #266, opened 2019-08-26
Duplicate routeDiagrams in RouteTableViewCell


Omarrasheed:
Find fix for updating tableview when delays occur. We currently just update the tableview from within the cell but because it's an asynchronous call and is run for each cell, we update the tableview way more often than we need to and prepareForReuse doesn't seem to get called as often. This causes overlaps in the routeDiagram and sometimes makes it impossible to read.

PR #337, opened 2019-10-27
Fix reuse configure function for routeTableViewCell

Omarrasheed:
## Overview

Remove the need for an extra nil check to remove from superView + remove constraints

## Changes Made

Changed constraints for routeDiagram to be `remakeConstraints`

## Test Coverage

Manual

## Related PRs or Issues (delete if not applicable)

#266

### DIFF
--- a/TCAT/Cells/RouteTableViewCell.swift
+++ b/TCAT/Cells/RouteTableViewCell.swift
@@ -140,8 +140,8 @@ class RouteTableViewCell: UITableViewCell {
             make.bottom.equalTo(routeDiagram.snp.top).offset(-routeDiagramTopOffset)
         }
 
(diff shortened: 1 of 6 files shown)


Podfile.lock
```diff
@@ -181,4 +181,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: b7d70eb804b087e359ae71cece52dfdbdb17f69c
 
-COCOAPODS: 1.8.4
+COCOAPODS: 1.8.3
```
young-k: am i misunderstanding something, but why would this number ever decrease?

Omarrasheed: Cuz im a bum. WIll update my cocoapods sir!

kevinchan159 (approved):
:shipit:

Issue #332, opened 2019-10-23
Scrap JSONFileManager


Omarrasheed:
We're not really using this and its bloating the hell out of our routeOptionsViewController and its only used for logs, which really doesn't end up getting used anyway. Lets kill it

PR #341, opened 2019-10-29
Remove JSONFileManager

lucyxubroad:
## Overview

We were using JSONFileManager so that we can send logs when users email us with feedback. However, the feature is largely unused and cluttered the app. This PR takes all that out. 

## Changes Made

 
Remove anywhere that JSONFIleManager is used and deleted the files! 

## Test Coverage

I ran the app as usual to see that things still work. Now when you send email to us from the info page, there will not be an attached log. 

## Related PRs or Issues (delete if not applicable)

#332 Scrap JSON Filemanager

</details>


danielVebman (approved):
Looks good to me.

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -560,17 +542,10 @@ class RouteOptionsViewController: UIViewController {
     }
 
     private func processRequest(result: Result<Response<RouteSectionsObject>>, requestURL: String, endPlace: Place) {
```
Omarrasheed: Pretty sure we can take out this whole function now that we don't need to log the requestURL. Also can we just take out anything pertaining to the requestURL

TCAT/Utilities/Time.swift
```diff
@@ -88,7 +88,7 @@ class Time {
         var dateComponents = Time.dateComponents(from: date!)
 
         // Modify date to have today's day, month & year
-        var todaysDateComponents = Time.dateComponents(from: Date())
+        let todaysDateComponents = Time.dateComponents(from: Date())
```
Omarrasheed, line 91: Love it

Omarrasheed (dismissed):
Just a small bit of cleanup left. Prob can take out the function that returns the requestURL within Endpoint and also can move the processRequest functions back so its less nested. Besides that LGTM. Just gonna request changes til its done

TCAT/Controllers/RouteOptionsViewController.swift
lucyxubroad: when you say "we can take out this whole function now", do you mean the processRequest function? Because I was under the impression that that is the function is actually loading our table results?

Omarrasheed: yes but i'd rather keep the logic in `processRequest` connected with the actual call to `.observe`. We can either move the 5 lines above `processRequest(...)` into `processRequest` or just move the code from `processRequest` into here. Now that I think of it tho, might be better to just move those 5 lines into `processRequest`
Screen Shot 2019-10-29 at 4 47 26 PM

danielVebman: A note about the code above: I'm fairly sure that as of swift 4.2 the guard in that image is equivalent to

    guard let self = self else { return }

i.e. drop the tick marks.

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -559,41 +533,40 @@ class RouteOptionsViewController: UIViewController {
         }
     }
 
-    private func processRequest(result: Result<Response<RouteSectionsObject>>, requestURL: String, endPlace: Place) {
-        JSONFileManager.shared.logURL(timestamp: Date(), urlName: "Route requestUrl", url: requestURL)
+    private func processRequest(start: Place, end: Place, time: Date, type: SearchType) {
+        if let result =  getRoutes(start: start, end: end, time: time, type: type) {
+            result.observe(with: { [weak self] result in
+                guard let `self` = self else { return }
```
danielVebman: Pretty sure you can drop the tick marks since Swift 4.2.

lucyxubroad: I hear ya! I think we have tick marks throughout the entire codebase right now? Do we wanna try to fix all of it at once?

TCAT/Controllers/RouteOptionsViewController.swift
lucyxubroad: @Omarrasheed I moved the function into our processRequest function so now we make the getRoutes call inside of process request and we make one call to process request from searchforroutes!

TCAT/Controllers/RouteOptionsViewController.swift
danielVebman: Honestly we could probably do it in one go with a find and replace of `` `self` `` –> `self` in the project. I guess the only reason not to do it right now/in this PR is if for whatever reason something goes wrong, but I don't think that's really possible.

danielVebman:
A note about the tick marks thing. I don't mean to be a pain about this, but: 

It's not that removing the tick marks now makes self strong. It's just syntactic sugar because the swift designers realized how ugly `` `self` = self `` looks. Even without the tick marks you are doing the same thing: defining a non-null variable called `self` if the `self` passed into the closure (which is *still* weak) is not `null`. It's the same thing as saying:

    guard let s = self else { return }
    s.resultsSection = [...]

`self` is really just an implicit variable. When you say let `self = self`, you're saying something like this:

    var banana: String = nil
    updateBanana(maybeUpdate: banana)
    guard let banana = banana else { return }
    // yay, banana is not null and we don't have to say stuff like `banana?.count`
    // similarly, we wouldn't have to say self?.resultsSection = [...]

I hope that made sense. I really like this kind of nitty gritty stuff. Sorry.

danielVebman:
About the merge conflict with `AllStopsTableViewController.swift`: I renamed that file in a previous PR, so it shouldn't exist anymore.

kevinchan159 (approved):
🚢

haiyingweng (approved):
lgtm!

TCAT/Controllers/RouteOptionsViewController.swift
```diff
@@ -559,41 +533,40 @@ class RouteOptionsViewController: UIViewController {
...
+                guard let self = self else { return }
+                DispatchQueue.main.async {
+                    switch result {
+                    case .value(let response):
 
-        switch result {
-        case .value(let response):
+                        // Parse sections of routes
+                        [response.data.fromStop, response.data.boardingSoon, response.data.walking]
+                            .forEach { (routeSection) in
```
young-k: @kevinchan159 is it convention to have `(routeSection)` here or just `routeSection`?

kevinchan159: `routeSection`

Issue #340, opened 2019-10-28
Make Transit Stable Again


Omarrasheed:
# CLEANUP CREW
Lots of work to be done but a lot of it shouldn't be too complex. Let's split up each of these tasks to a PR so that we can avoid as many merge conflicts as possible while others r working.

*Before working on a task, please tag yourself at the end of the task description so we're not doing much overlapping work*

- [x] **[Daniel]** Look into removing `UnwindAllStopsTVCDelegate`
- [x] **[Daniel]** No clue what these lines are in `AllStopsViewController` under UITableViewDelegate extension. Pretty sure we can remove them or move them into `PlaceTableViewCell`. [Reference](https://user-images.githubusercontent.com/36868927/67649301-e0ed0980-f90e-11e9-888a-63dbaaeb9f37.png) 
- [ ] Lets just wrap this whole thing in a chained if let so we can safely unwrap section and place (and sortedKeys[indexPath.section]??) and so we don’t have this weird print statement within the guard. [Reference](https://user-images.githubusercontent.com/36868927/67649326-0548e600-f90f-11e9-8cc1-e165d0fd4206.png)
- [ ] Can we look into removing this code from HomeMapViewController? We don’t really have a loading state for the map like ever. [Reference](https://user-images.githubusercontent.com/36868927/67649600-12b2a000-f910-11e9-8674-e637d5912e82.png)
- [ ] In setting up the InformationTableViewController, we have this random function that changes the text color based on the row & section number but I feel like this should be based on the section name, considering these come from shared constants. This would prob look wack if we decide to change the ordering at some point down the line. [Reference](https://user-images.githubusercontent.com/36868927/67649699-85bc1680-f910-11e9-88a1-5306aee5f368.png)
- [ ] Our searchViewController also seems pretty jank. Just kinda lookin thru that and seeing 
- [ ] Arbitrary delay being used in ServiceAlertsViewController??? [Reference](https://user-images.githubusercontent.com/36868927/67649617-2fe76e80-f910-11e9-8a73-542b29a02bcc.png)
- [ ] I feel uncomfortable about having this as an attribute of the BusLocation model. If we can move this to a function of a BusLocation instance or even take it out completely that’d be better. [Reference](https://user-images.githubusercontent.com/36868927/67649900-3d512880-f911-11e9-8160-7a3a1d5d3173.png)
- [ ] BusPath object should probably be in views and not in the models folder
- [x] Confused about our LegacyModels.swift file??
[...]

PR #343, opened 2019-10-31
[MTSA] Remove migration code

danielVebman:
## Overview

For old users (before v1.2), places were stored using different data structures.

## Changes Made

Removed migration code in AppDelegate and deleted legacy models.

## Test Coverage

Ran the app to make sure nothing is broken. Looking at how favorites are handled, it seems that for the few (if any) users who are still on the super old version, their favorites will just be overwritten when they tried to read them.

## Related PRs or Issues

Part of the continuing series, *Make Transit Stable Again #340.*


young-k (approved):
i love deleting code, send it

Omarrasheed (approved):
Looks dope. Again lets j make sure Siri shortcuts still work

kevinchan159 (approved):
👏👏

PR #348, opened 2019-11-11
[MTSA][Refactor reachability][Part 1/2] Add ReachabilityManager

danielVebman:
## Overview

Currently, all reachability updates are passed down to child view controllers through a gross series of delegate method calls from the root UINavigationController. This PR and the next one replace that rather serpentine system with a singleton `ReachabilityManager` than any object can subscribe to for reachability updates.

## Changes Made

Added a static singleton reachability manager that holds weak references to listeners so they are automatically removed once deinitialized. 

## Test Coverage

Added a test to make sure that the weak references stuff was actually working.

## Next Steps

Refactor all references to Reachability so that they instead add a closure to the `ReachabilityManager`.

## Related PRs or Issues

Part of the continuing series, Make Transit Stable Again #340.

I think you guys get the point by now. Maybe I'll start omitting that line ^.

### DIFF
--- a/TCAT.xcodeproj/project.pbxproj
+++ b/TCAT.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		0645397123789C0400A4ED81 /* RetainTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0645397023789C0400A4ED81 /* RetainTest.swift */; };
 		069676ED23738C1F004F7152 /* ReachabilityManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = 069676EC23738C1F004F7152 /* ReachabilityManager.swift */; };
 		06C4EFD9236A755300D97117 /* AddFavoriteTableViewCell.swift in Sources */ = {isa = PBXBuildFile; fileRef = 06C4EFD8236A755300D97117 /* AddFavoriteTableViewCell.swift */; };
 		222BDE16215C1CEE0040DD93 /* WhatsNewHeaderView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 222BDE15215C1CEE0040DD93 /* WhatsNewHeaderView.swift */; };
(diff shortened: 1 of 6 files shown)


TCAT.xcodeproj/project.pbxproj
```diff
@@ -1403,7 +1411,7 @@
 				BUNDLE_LOADER = "$(TEST_HOST)";
 				CURRENT_PROJECT_VERSION = 47;
 				INFOPLIST_FILE = TCATTests/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
```
kevinchan159, line 1414: are we sure we want to do this?

TCATTests/RetainTest.swift
```diff
@@ -0,0 +1,55 @@
...
+import Foundation
+import UIKit
+
+class TestRetainVC: UIViewController {
+    let v = UIView()
+    
+    init() {
+        super.init(nibName: nil, bundle: nil)
+        
+        TestManager.shared.addListener(self) { [weak self] (s) in
```
kevinchan159: remove parens around `(s)`

TCATTests/RetainTest.swift
```diff
@@ -0,0 +1,55 @@
...
+//  TCATTests
+//
+//  Created by Daniel Vebman on 11/10/19.
+//  Copyright © 2019 cuappdev. All rights reserved.
+//
+
+import Foundation
+import UIKit
+
+class TestRetainVC: UIViewController {
```
kevinchan159, line 12: newline after 12

TCATTests/RetainTest.swift
```diff
@@ -0,0 +1,55 @@
...
+        TestManager.shared.addListener(self) { [weak self] (s) in
+            self?.v.backgroundColor = .red
+            print("Recevied " + s)
+        }
+    }
+    
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+}
```
kevinchan159, line 29: newline before here

kevinchan159 (approved):
looks great!!! we've also never ever really created // used Tests before so it's great to see that you added some 🤩

Omarrasheed (approved):
Ah tests we don't do that here
jk jk LGTM

TCAT.xcodeproj/project.pbxproj
danielVebman: My tests wouldn't compile otherwise.

PR #349, opened 2019-11-15
Update NotificationBannerSwift version

haiyingweng:
## Overview

There were some iOS13 problems with `StatusBarNotificationBanner`
Updated version number for `NotificationBannerSwift` cocoapod

## Changes Made

Updated version number for `NotificationBannerSwift` cocoapod with iOS13 fixes

## Test Coverage

Ran on simulator


lucyxubroad (approved):
if we only have one notification on screen at a time, do we still wanna follow this syntax of queue: ----- ? I'm not very familiar with NotificationBannerSwift but it looks good! and good catch on the bug

kevinchan159 (approved):
🚢

PR #361, opened 2020-08-28
Fix date string

whoiswillma:
## Overview

Change date string to match value returned by backend.


Omarrasheed (changes requested):
Good catch, was this a mistake caused by a previous pr or is this a change to the data formatting from backend? 
If the latter, then we should fix this on backend as that should help fix old clients

We should potentially look into parsing dates regardless of their format, I feel like that should exist somewhere. That orrrr we enforce a format on backend (regardless of what's in the gtfs data)

Cc @alanna-zhou

PR #365, opened 2020-09-12
Zoom in on viewDidAppear

Omarrasheed:
## Overview

Fix the bug w map not zooming to proper bounds

## Changes Made

Essentially the zoom function was being called on drawer being moved, but because the drawer is initialized before the map fully loads, it tries to zoom the map before it exists and thus no-ops. This should handle that edge case

## Test Coverage

Peep screenshot

## Next Steps (delete if not applicable)

Fix more bugs i guess

## Screenshots (delete if not applicable)

ezgif com-video-to-gif


ywsang (approved):
<3

PR #366, opened 2020-09-12
Remove unnecessary fields from Route model

ywsang:
## Overview

Clean up the `Route` model

## Changes Made

Removed `timeUntilDeparture`, `startCoords`, `endCoords`, `numberOfTransfers`, `boundingBox`.

## Test Coverage

Ran app with several route searches

## Screenshots 

 IMG_6189

### DIFF
--- a/TCAT/Model/Route.swift
+++ b/TCAT/Model/Route.swift
@@ -54,25 +54,8 @@ class Route: NSObject, Codable {
     /// A unique identifier for the route
     var routeId: String
 
-    /// The amount of time from now until the departure
-    var timeUntilDeparture: DateComponents {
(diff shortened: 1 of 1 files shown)


Omarrasheed (approved):
Ur the goat. Just curious, did u get a chance to go thru the fields in `Direction` class as well?

jk i assume u went thru it cuz its in the doc

ywsang:
> Ur the goat. Just curious, did u get a chance to go thru the fields in `Direction` class as well?
> 
> jk i assume u went thru it cuz its in the doc

yes sir
we use all of them but the types could probably be changed
like `tripIdentifiers` is an array and prob doesn't have to be

PR #369, opened 2023-12-27
Fixed deprecation issues

vinnie4k:
## Overview

Fixed many deprecation issues to prepare Transit for revival.

## Changes Made

- Updated SwiftLint rules and build script phase.
- Increased minimum iOS version to 15.0 from 11.0.
- Updated Firebase Analytics.
- Updated CocoaPods.
- Removed UITests in Targets since these weren’t being used.
    - Kept UnitTests though in case we need them.
- Fixed linting issues.
- Removed storyboard files (except Launchscreen).
- Removed TodayExtension since this is outdated.
    - We can configure widgets in a future update.
- Removed fastlane automation system.

## Test Coverage

- Ran on both dev and prod servers.
- Ran on physical device with the latest iOS version (17.0).

## Next Steps

- I realized that our backend has a “v3” but we are currently using “v2”. Will need to discuss this with pod lead and backend lead.
- Should be ready for development.


antoinztte (approved):
LGTM

PR #378, opened 2024-04-10
Add Apollo Client

vinnie4k:
## Overview

Added the Apollo iOS Client to the codebase to integrate with Uplift’s GraphQL backend.

### Apollo

- The Codegen CLI looks at the configuration JSON files (in our secrets) as well as our `.graphql` files to generate the API.
- There is a build script that runs the Codegen CLI executable based on the currently selected schema (DEBUG or RELEASE).
- The Swift code defined inside of `UpliftAPI` has been configured with the `internal` access modifier. Because I have configured it to be used directly within the target instead of a library, we will need to prepend `UpliftAPI.` (with the dot) to access anything generated by Apollo, which is nested inside of an enumeration called `UpliftAPI`.

### Networking

- Created `ApolloNetwork.swift` to contain code that accesses the Apollo Client.
- Created `Publishers.swift` that extends the `Publishers` class defined in the Combine framework.
    - This file contains two structs (one for queries, one for mutations) that is used for configuring the publishers used.
    - These configurations are very flexible and have support for optimization with caching, concurrency, etc:
        - `cachePolicy`: A cache policy that specifies when results should be fetched from the server and when data should be loaded from the local cache. Defaults to `.default`.
        - `contextIdentifier?`: A unique identifier for this request, to help with deduping cache hits for watchers. Defaults to `nil`.
        - `context?`: A context that is being passed through the request chain. Defaults to `nil`.
        - `queue`: A dispatch queue on which the result handler will be called. Defaults to the main queue.
        - `publishResultToStore`: If `true`, this will publish the result returned from the operation to the cache store. Defaults to `true`. For mutations only.
    - This file also defines a class called `ApolloMutationSubscription` that processes the subscription that is received by the publishers.
- Created `ApolloClientProtocol.swift` that extends the `ApolloClientProtocol` class defined in the Apollo and ApolloAPI libraries.
    - This file contains two functions (one for queries, one for mutations) that returns the Combine publishers defined in `Publishers.swift`.

### Other Changes

- Updated `README.md` with the up-to-date instructions for setting up the codebase.
- Temporarily removed the `AppDevAnnouncements` package because it is currently being refactored. See added `// TODO` comments in the codebase.
- Removed the TCATTest target/testing suite because it is outdated.
- Removed Siri extension because it is outdated.
- Updated our secrets/environment variables with Eatery and Uplift’s backend server URL.
    - Note that Transit’s URL now has an `https://` prepended. To account for this, the current code removes `https://` from the URL. The reason for this change is because we will be refactoring networking in the codebase soon.
- Reorganized the project directory since there were files that were not properly in their folder on Finder, but were organized correctly in the project.
- Minor syntax changes to the codebase.

## Test Coverage

- Tested on the simulator and a physical device for mainly UI issues.

## Next Steps

- Since everything works locally in development, I will need to see if Xcode Cloud works with these new changes.
- Update our Xcode Cloud workflow, Slack channels, and secrets manager with the new secrets and environment variables.
- Write Uplift models to be used within the codebase.
- Refactor networking and add AppDev Announcements once they’re ready.


daniel-chuang (commented):
My goat

PR #385, opened 2024-06-02
Fixed small bug with date picker

cindy-x-liang:
## Overview

Made some fixes with the UI of the date picker

### Change 1

-Centered the date picker

### Change 2

- Made it so that you can only choose 1 of the 3 options, previously you could choose Leave Now and either Leave at/Arrive By

## Screenshots (optional)
image

### DIFF
--- a/TCAT/Views/DatePickerView.swift
+++ b/TCAT/Views/DatePickerView.swift
@@ -84,11 +84,6 @@ class DatePickerView: UIView {
         styleSegmentedControl(timeTypeSegmentedControl)
         setSegmentedControlOptions(timeTypeSegmentedControl, options: [leaveNowElement.title,leaveAtElement.title, arriveByElement.title])
         timeTypeSegmentedControl.selectedSegmentIndex = leaveNowElement.index
-        timeTypeSegmentedControl.addTarget(
-            self,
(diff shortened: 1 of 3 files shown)


TCAT/Views/DatePickerView.swift
```diff
@@ -201,6 +180,10 @@ class DatePickerView: UIView {
             default:
                 break
             }
+        }else {
```
rs929: Add a space between } and "else"

TCAT/Views/DatePickerView.swift
```diff
@@ -213,9 +196,6 @@ class DatePickerView: UIView {
     // MARK: - Segment Controls
 
     @objc private func timeTypeSegmentedControlValueChanged(segmentControl: UISegmentedControl) {
-        if timeTypeSegmentedControl.selectedSegmentIndex == arriveByElement.index {
-            leaveNowSegmentedControl.selectedSegmentIndex = UISegmentedControl.noSegment
-        }
     }
```
rs929: Is this function still being used? It seems like there is no body. If it's not being used just remove it.

rs929 (approved):
Hey Cindy, great job with the fix! Just a few small style issues, please fix them before you merge.

Issue #389, opened 2024-07-28
Search results should be filtered by distance


cindy-x-liang:
The closest target is not the first search result
![IMG_6521](https://github.com/user-attachments/assets/b92d1ee8-ca85-415b-89f5-f54f4c41dda1)

PR #391, opened 2024-08-05
add route filtering

cindy-x-liang:
## Overview

An attempt to filter search results by distance. Previously when you search Target the first result is one that is out of range.

### Change 1

Added a sort function in the search manager

## Test Coverage

I tested with Target and some other popular locations in Ithaca. (Trader Joe's, Four Seasons, De Tasty) 
However the search is kindy finicky I think due to caching, would be great if someone else could test before merge

## Related PRs or Issues (optional)

Issues: #389 

## Screenshots (optional)
Before:
image

After:
IMG_1964


rs929 (approved):
Great work! LGTM 🚀

PR #402, opened 2025-02-12
Visibility of Google Maps Icon

angelinaa-chen:
## Overview

The tiniest change ever but the Google Maps logo used to be hid behind the Favorites Bar and now we can see it! Also the original link to our apps wasn't working on Simulator (pretty sure Apple switched from itunes.apple to apps.apple) so I just went ahead and updated that too.

image


Podfile.lock
```diff
@@ -239,4 +239,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: af336d88f53594af448d02dc18637c2b6ebe685e
 
-COCOAPODS: 1.15.0
+COCOAPODS: 1.16.2
```
rs929: Just double checking that you intended to update Cocoapods for this repo, otherwise I would try to remove this change lol

rs929 (approved):
Great work! Just wanna double check that you intended to update CocoaPods

Podfile.lock
angelinaa-chen: Ohh I think that updated automatically, but I'll change it back just to be safe!

JaysonHahn (approved):
Looks good to me! Look into though eventually if the google logo could move up and down with the sheet.

PR #409, opened 2025-10-10
Backup/before rollback

cindy-x-liang:
implementing notifs, building off of jayson's


TCAT/Cells/NotificationToggleTableViewCell.swift
```diff
@@ -77,23 +80,85 @@ class NotificationToggleTableViewCell: UITableViewCell {
...
 
     @objc func switchValueChanged() {
-        if notificationSwitch.isOn {
+
+        let isOn = notificationSwitch.isOn
+        
+        setToggle(isOn, for: type, tripId: tripId)
+
+  
+
```
MrPeterss: nit: can we remove this whitespace?

TCAT/Controllers/RouteDetail+ContentViewController.swift
```diff
@@ -205,7 +206,9 @@ class RouteDetailContentViewController: UIViewController {
                 guard let self = self else { return }
 
                 if case .failure(let error) = completion {
+
                     self.printClass(context: "\(#function) error", message: error.errorDescription)
+
```
MrPeterss: can we remove these two newlines?

TCAT/Controllers/RouteDetailDrawerViewController+Extensions.swift
```diff
@@ -206,11 +206,36 @@ extension RouteDetailDrawerViewController: UITableViewDataSource {
...
+                isFirst: false,
+                delegate: self,
+                startTime: startTime,
+                tripId: tripId,
+                stopId: stopId
             )
+
+            // Make sure the visual switch matches your persisted state
+            // If you haven’t added an `isOn` parameter to configure, set it directly:
+            // (Alternatively, add an `isOn` parameter to `configure` and set inside that method.)
```
MrPeterss: can we remove this comment?

TCAT/Supporting/TransitEnvironment.swift
```diff
@@ -28,6 +28,9 @@ enum TransitEnvironment {
         static let announcementsHost = "ANNOUNCEMENTS_HOST"
         static let announcementsPath = "ANNOUNCEMENTS_PATH"
         static let announcementsScheme = "ANNOUNCEMENTS_SCHEME"
+        
+        // TODO: Remove once the Notifications moves to prod
+        static let devTransitURL = "TRANSIT_DEV_URL"
```
MrPeterss, line 33: We should be careful of making sure to remove this sooner rather than later

MrPeterss (approved):
Looks great! just a couple of nits

TCAT/Supporting/TransitEnvironment.swift
cindy-x-liang: will do!

TCAT/Managers/TransitNotificationSubscriber.swift
```diff
@@ -0,0 +1,114 @@
...
+                receiveCompletion: { completion in
+                    if case .failure(let error) = completion {
+                        print("Failed to subscribe to departure notification: \(error)")
+                    }
+                },
+                receiveValue: { success in
+                    print("Departure notification subscription success: \(success)")
+                }
+            )
+            .store(in: &self!.cancellables)
```
AsenKimO: idk how I feel abt force unwrapping here. It feels wrong but ig it will never be an issue since this function (and the ones below as well) will be called through the `shared` singleton. Wdyt @MrPeterss

TCAT/Services/Transit/TransitProvider.swift
```diff
@@ -16,7 +16,11 @@ enum TransitProvider {
     case applePlaces(ApplePlacesBody)
     case appleSearch(SearchResultsBody)
     case busLocations(GetBusLocationsBody)
+    case cancelDelayNotification(DelayNotificationBody)
+    case cancelDepartureNotification(DepartureNotificationBody)
     case delay(GetDelayBody)
+    case delayNotification(DelayNotificationBody)
+    case departueNotification(DepartureNotificationBody)
```
AsenKimO: departure misspelled

TCAT/Services/Transit/TransitProvider.swift
```diff
@@ -25,7 +29,15 @@ extension TransitProvider: ApiEndpoint {
 
     /// Base URL string for the transit API.
     var baseURLString: String {
-        return TransitEnvironment.transitURL
+//        return TransitEnvironment.transitURL
+        // TODO: Remove once the Notifications moves to prod
```
AsenKimO, line 33: I think peter mentioned removing `devTransitURL` sooner than later. What the plan for this? Just removing this as soon as notifs are available on prod?

AsenKimO (changes requested):
Just some minor comments. Otherwise, LGTM 🔥

TCAT/Managers/TransitNotificationSubscriber.swift
MrPeterss: good catch asen yeah lets not force unwrap this. On line 19 just add a guard for self to that existing guard and then you shouldn't have any issues if you remove the `!`

`guard let token = token, let self, ...`

AsenKimO (approved):

TCAT/Services/Transit/TransitProvider.swift
cindy-x-liang: Yes! Also - it's mapped right now to my local machine but I think we should first push the backend changes I made to the testflight backend server, change the TRANSIT_DEV_URL to the testflight URL, and then when everything looks good there push everything to prod and get rid of this code.

